PR #201, opened 2026-04-11
chore: migrate spec-sync to v4.0.0 directory structure

corvid-agent:
## Summary
- Migrates from flat `specsync.json` + `specsync-registry.toml` to `.specsync/` directory structure
- `.specsync/config.toml`, `.specsync/registry.toml`, `.specsync/version`
- Fixes `.gitignore` to track `archive/` and `hashes.json` (part of v4 lifecycle, should be version-controlled)
- Updates the `specsync migrate` template so new projects also get the correct gitignore
- Adds `hashes.json` to the repo

## Changes
- `.specsync/` directory with config.toml, registry.toml, version, .gitignore
- `src/commands/migrate.rs` — updated gitignore template to not ignore archive/ or hashes.json
- `tests/integration.rs` — updated test to verify neither archive/ nor hashes.json are gitignored
- Removes old `specsync.json` and moves `specsync-registry.toml` into `.specsync/`

Supersedes #202.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #202, opened 2026-04-11
fix: stop gitignoring .specsync/archive/ directory

corvid-agent:
## Summary
- Removes `archive/` from `.specsync/.gitignore` — archived/retired specs should be version-controlled, not hidden from git
- Updates the migrate command template so new projects also get the fix
- If users don't want archived specs, they can delete them explicitly — silently gitignoring loses history

## Changes
- `.specsync/.gitignore` — removed `archive/` entry, updated comments
- `src/commands/migrate.rs` — same fix in the gitignore template written by `specsync migrate`


corvid-agent:
Consolidated into #201 which now includes all the gitignore fixes (both archive/ and hashes.json) plus the migrate template update.

PR #297, opened 2026-07-02
Fix: merge no longer corrupts specs by dropping frontmatter fences (Critical: data loss)

0xLeif:
## Summary
Fixes a **Critical data-loss bug** found while dogfooding, and — after adversarial review probed deeper — several related **pre-existing** content-loss bugs in `specsync merge` (the flagship auto-resolve-conflicts command). The unifying fix: **only auto-resolve a conflict hunk when it can be merged losslessly; otherwise leave it for manual resolution with the file untouched.**

## What was corrupting / losing content (all now closed)
1. **Dropped frontmatter fences** — a hunk holding the whole `---…---` block resolved to loose/doubled/empty frontmatter → `check` says "Frontmatter invalid".
2. **Deleted spec body** — an add/add hunk spanning the closing `---` + body was rebuilt from parsed fields alone, deleting the body.
3. **Separator-before-fence** — a blank/whitespace/comment line between the clean opening `---` and a fence-carrying hunk produced an empty-frontmatter spec.
4. **Pre-`##` body hunk** — a `# Notes`/intro conflict (no `##` heading) was routed to the frontmatter field resolver, dropping bullet lists and prose.
5. **Impure Change Log hunk** — a changelog conflict spanning into the next section deleted that section (the row merge kept only `|`-rows).

## Fix
- **No fence reconstruction.** The field resolver runs only for *pure interior field conflicts* (`is_frontmatter_only`: every line a `key: value`/`- item`) while genuinely inside the frontmatter (`in_frontmatter`: <2 fence lines emitted). Any hunk carrying a `---`, heading, or body → Manual.
- **All row merges gated on `is_pure_table_rows`** (changelog + generic table), so a hunk that swallowed prose/section → Manual.
- **Shape-independent backstop** — if the input had frontmatter but the output doesn't parse or lost its `module`, downgrade to Manual. `merge` writes only on `Resolved`, so a risky resolution never overwrites the original.

## Not in scope (deferred, by-design)
- Generic table resolver unions rows by first cell, so two distinct rows sharing a first cell (e.g. a class name with two method rows) collapse. Union-by-key is intended for symbol-keyed API tables; tracked as a follow-up.
- Scalar "theirs wins" for a genuine frontmatter field conflict (a version/status difference resolves to the incoming side) — not data loss.

## Verified
- All prior repros (fence-in-hunk, body-swallow, separator, pre-heading body, impure changelog) fall to Manual with the file byte-identical (markers + body + fields preserved), confirmed end-to-end on the real CLI.
- Common cases still auto-resolve: interior `version` bump → valid frontmatter (theirs wins); pure changelog rows → unioned chronologically.
- 14 merge unit tests + 2 integration tests; full suite (660 unit + 151 integration), self-check 60/60 at 100%. CI clippy (bin) + fmt clean.

### DIFF
--- a/src/merge.rs
+++ b/src/merge.rs
@@ -380,29 +380,45 @@ fn resolve_table_conflict(ours: &str, theirs: &str) -> Resolution {
     Resolution::Auto(format!("{merged}\n"))
 }
 
-/// Whether a conflict-hunk side is ONLY interior frontmatter fields — every
-/// non-blank line is a YAML `key: value` (spaceless key) or `- item` list line.
+/// Whether a conflict-hunk side is ONLY interior frontmatter fields that
+/// [`parse_yaml_fields`] can round-trip WITHOUT dropping anything.
 ///
-/// Deliberately strict: a `---` fence, a `##` heading, a table row, or any prose
-/// line disqualifies it. We only auto-merge a frontmatter conflict when BOTH sides
-/// are pure fields with the `---` fences left in the surrounding clean regions
-/// (the common "two branches bumped `version`" case). Reconstructing fences that a
-/// hunk swallowed is not attempted — a hunk carrying a fence or body is left for
-/// manual resolution rather than risk deleting body content or emitting a broken
-/// or doubled fence.
+/// It must mirror the parser's keep/drop rules exactly, or a hunk it accepts could
+/// still lose content on merge:
+/// - a `---` fence, `##` heading, table row, or prose line ⇒ reject (not a field);
+/// - a `- item` list line is kept by the parser only while it follows a key with an
+///   empty/`[]` value (a list key). A `- item` before any such key — its owning key
+///   sits in the surrounding clean region — is silently dropped by the parser, so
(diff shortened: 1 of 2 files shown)


0xGaspar (approved):
Strong fix, and the right shape for it. Making the field resolver fire only for pure interior-field hunks (`is_frontmatter_only` + `in_frontmatter`), gating every row merge on `is_pure_table_rows`, and refusing to reconstruct swallowed `---` fences all attack the root cause — reconstructing structure from parsed fields alone — rather than patching individual repros. The `module`-present backstop combined with `merge` writing only on `Resolved` means a risky resolution leaves the original byte-identical, which is exactly the fail-safe you want for a data-loss class. `in_frontmatter` via "<2 fence lines emitted" is sound because a hunk carrying a fence is already rejected upstream, so the count can't be fooled by a body/horizontal-rule `---`. Coverage (14 unit + 2 integration, every prior repro verified falling to Manual) is convincing, and deferring union-by-first-cell table collapse is a reasonable scope line since it's not new data loss. LGTM.

PR #305, opened 2026-07-03
Fix: parse multi-line TOML arrays instead of corrupting the config (data loss)

0xLeif:
## The bug (surfaced by the C1 migrate review · data loss in the primary config format)
`load_toml_config` parses line-by-line. A formatter-style multi-line array:
```toml
source_dirs = [
  "src",
  "lib",
]
```
was read as the value `[` alone → `parse_toml_string_array("[")` fell through to **`["["]`**, and the item lines (no `=`) were silently dropped. The user's real `source_dirs`/`exclude_patterns`/etc. became a single bogus `[` entry — so specsync scanned a nonexistent directory named `[`, found 0 source files, and reported vacuous 100% coverage / a green build.

**Worse:** `migrate` reads *then rewrites* the config, so it **persisted the corruption**, destroying a valid `.specsync.toml`. Every array key was affected.

## The fix
- **Accumulate continuation lines** when a value opens `[` without closing on the same line, until the array closes — multi-line arrays now parse to their real entries.
- **`parse_toml_string_array` uses the span between the first `[` and the LAST `]`**, tolerating a trailing comment and the joined multi-line text (also fixes the single-line `[...] # comment` case, and preserves a `]` inside a quoted glob like `**/[abc]/**`).
- **Strip inline `#` comments (quote-aware)** from array lines, reusing the existing `strip_inline_comment`; whole-line comments and blanks inside an array are skipped. **Scalar value parsing is byte-for-byte unchanged.**

## Verified
```
$ specsync migrate --no-backup   # legacy multi-line .specsync.toml
→ .specsync/config.toml:  source_dirs = ["src", "lib"]   # was ["["]
```

Tests: multi-line arrays (global + `[lifecycle]` section), comment-inside-array, empty multi-line array, key-after-array still parsed, trailing comment, glob-with-brackets, plus an integration test that `migrate` preserves a multi-line `.specsync.toml`. Full suite 685 + 160, fmt / clippy (bin) / self-check 100%.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -690,12 +690,14 @@ fn load_toml_config(config_path: &Path, root: &Path) -> SpecSyncConfig {
                         lines.next(); // consume blank / whole-line comment in array
                         continue;
                     }
-                    // Array items are quoted strings (or a nested `[`); the close is
-                    // `]`. A continuation line that is none of these means the `]`
-                    // was omitted — stop WITHOUT consuming it so the following
-                    // key/section still parses, bounding a malformed array's damage
-                    // to its own key rather than swallowing the rest of the file.
-                    if !cont.starts_with('"') && !cont.starts_with(']') && !cont.starts_with('[') {
+                    // Array items are quoted strings; the close is `]`. A
+                    // continuation line that is neither means the `]` was omitted —
+                    // stop WITHOUT consuming it so the following key/section still
+                    // parses, bounding a malformed array's damage to its own key
+                    // rather than swallowing the rest of the file. (These configs
+                    // never use nested arrays, so a leading `[` is a following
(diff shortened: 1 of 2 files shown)


Kyntrin (changes requested):
One blocking repo-process issue remains before merge:

- `src/config.rs` changed config-module behavior, but `specs/config/config.spec.md` was not updated. Per the repo instructions, please increment the config spec `version` and add a Change Log entry for the multi-line TOML array parsing change.

The previous quote/bracket handling concern appears addressed in the new commit via `find_toml_array_close` and the added multi-line bracket-glob coverage. I also checked `git diff --check` locally and it passed.

PR #306, opened 2026-07-03
Fix: hooks install --claude-code-hook deep-merges, preserving user hooks (High)

0xLeif:
## The bug (dogfooding finding H3 · High · data loss)
`install_claude_code_hook` merged specsync's hook into an existing `.claude/settings.json` with a blind:
```rust
obj.insert("hooks".to_string(), hooks.clone());   // overwrites the WHOLE hooks object
```
Anyone who already had their own `PreToolUse` / `PostToolUse` / other Claude Code hooks configured **lost them silently** the moment they ran `specsync hooks install --claude-code-hook`.

## The fix
Replace the blind insert with a per-event **deep merge** (`merge_claude_code_hooks`):
- an event the user doesn't have → added;
- an event they do have (an array of matcher groups) → specsync's group is **appended** to their array;
- other events (`PreToolUse`, …) and unrelated top-level settings (`permissions`, `model`, …) are untouched;
- a `hooks` value present but not an object (malformed) is reset to a working object rather than merged into a non-map;
- the existing `contains("specsync")` idempotency guard still prevents a second install from double-appending.

## Verified
Installing into a settings.json that already has `permissions` + a user `PreToolUse: [audit.sh]`:
- `permissions` preserved ✓
- `PreToolUse` (audit.sh) preserved ✓
- specsync's `PostToolUse` appended ✓

Tests: `install_claude_code_hook_preserves_existing_user_hooks` (unit) + a CLI integration test. Full suite 680 + 160, fmt / clippy (bin) / self-check 100%.

## Note
Uninstall already declines to auto-edit settings.json (it prints "remove manually"), so it is non-destructive; this PR is scoped to the install clobber. Auto-removing just specsync's group on uninstall would be a reasonable follow-up.

### DIFF
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -499,9 +499,9 @@ fn install_claude_code_hook(root: &Path) -> Result<bool, String> {
             .expect("built-in hook template is valid JSON");
 
         if let Some(obj) = parsed.as_object_mut()
-            && let Some(hooks) = hook_value.get("hooks")
+            && let Some(new_hooks) = hook_value.get("hooks").and_then(|h| h.as_object())
         {
-            obj.insert("hooks".to_string(), hooks.clone());
+            merge_claude_code_hooks(obj, new_hooks);
         }
 
         let new_content = serde_json::to_string_pretty(&parsed)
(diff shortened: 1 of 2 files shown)


Kyntrin (changes requested):
Two blocking issues before merge:

1. `src/hooks.rs` non-object settings root silently reports success

If `.claude/settings.json` is valid JSON but the root is not an object, for example `[]` or `null`, install skips `merge_claude_code_hooks`, writes the same value back, and returns `Ok(true)`. The CLI can report a successful install even though no specsync hook was added. Please return an error for a non-object settings root, or explicitly replace it if that is the intended recovery behavior, and add a regression test.

2. Hooks spec was not updated

`src/hooks.rs` behavior changed, but `specs/hooks/hooks.spec.md` was not updated. Per repo instructions, please bump the spec version and add a Change Log entry. It would also be useful to document the invariant that Claude Code hook install deep-merges user hooks instead of replacing the `hooks` object.

Verification performed locally on `pr-306-review`:
- `git diff --check origin/main...HEAD` passed
- `cargo test install_claude_code_hook_preserves_existing_user_hooks` passed
- `cargo test --test integration hooks_install_claude_code_hook_preserves_user_settings` passed
- `./target/debug/specsync check --strict` passed, 60 specs checked

PR #315, opened 2026-07-04
Fix: `diff` fails loud on a bad base ref instead of reporting "no changes"

0xLeif:
Medium-tier audit finding (#13) — a **fail-open** that violates the project's "fail loud" mandate.

## Problem

`cmd_diff` only handled the case where `git diff` fails to **spawn** (`Err` → exit 1). It never inspected `output.status`. Git *runs* but exits non-zero on error (bad base ref, not a git repository) with **empty stdout** — which the "no files changed" path reported as `No files changed since <base>` and **exited 0**.

So a failed comparison is silently reported as "no drift". In CI, `specsync diff --base <bad-ref>` — even with `--strict` — would **green-light a PR that was never actually diffed**.

## Reproduced

```
$ specsync diff --base totally-bogus-ref      # git exits 128, empty stdout
No files changed since totally-bogus-ref
EXIT=0                                          # ← fail-open
```

After the fix:

```
$ specsync diff --base totally-bogus-ref
git diff failed for base ref `totally-bogus-ref`: fatal: ambiguous argument ...
  Is `totally-bogus-ref` a valid revision, and is this a git repository?
EXIT=1                                          # ← fail-loud
```

A valid base (with or without changes) is unaffected.

## Fix

After `.output()`, check `output.status.success()`; on failure, surface git's stderr, name the base ref, and `exit(1)`.

## Spec contradiction resolved

The spec already carried a contradiction this fix settles:
- `cmd_diff.spec.md` **Error Cases**: "`git diff` fails (bad ref) → Exits with code 1" ✅ (correct intent)
- `testing.md` **Regression Matrix**: documented the *actual buggy* behavior — "treats it as no changes … add a fixture **before adding hard-failure semantics**" ❌

Code now matches the Error Cases. Added invariant #6, updated the Regression Matrix + Automated Coverage to cite the new test.

## Tests

- `diff_bad_base_ref_fails_loud` (integration): a bogus `--base` exits non-zero, names the ref on stderr, and never prints "No files changed".
- **732 unit + 169 integration**, `cargo fmt --check` clean, `cargo clippy --bin specsync -- -D warnings` clean, self-check 60/60 at 100%.

### DIFF
--- a/specs/cmd_diff/cmd_diff.spec.md
+++ b/specs/cmd_diff/cmd_diff.spec.md
@@ -32,7 +32,7 @@ Implements the `specsync diff` command — shows which specs are affected by sou
 
 1. Uses `git diff --name-only {base}` to get the list of changed files
 2. Only files matching `config.source_extensions` are considered
-3. Export deltas are computed by comparing current exports against base-ref exports (via `git show`)
+3. Export deltas compare each affected spec's currently-exported symbols (parsed from the working-tree source files) against the symbols it documents: `new_exports` are current exports the spec does not document, `removed_exports` are documented symbols no longer exported (there is no base-ref content fetch)
 4. Changed files not covered by any spec are listed separately
 5. Text output delegates to `output::print_diff_markdown`
 6. A non-zero `git diff` exit (bad base ref, not a git repository) fails loud with exit code 1 — it is never treated as an empty change set ("No files changed"), so a failed comparison cannot silently pass in CI
--- a/specs/cmd_diff/cmd_diff.spec.md
+++ b/specs/cmd_diff/cmd_diff.spec.md
@@ -35,6 +35,7 @@ Implements the `specsync diff` command — shows which specs are affected by sou
 3. Export deltas are computed by comparing current exports against base-ref exports (via `git show`)
 4. Changed files not covered by any spec are listed separately
 5. Text output delegates to `output::print_diff_markdown`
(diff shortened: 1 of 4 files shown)


0xLeif:
**Adversarial review: READY_WITH_NITS, 93%, zero blocking.** Verified empirically that `git diff --name-only` (no `--exit-code`) exits 0 on a valid ref whether or not there are changes — so there is **no false-failure risk**; the new gate fires only on genuine git errors. `diff.rs` has exactly one git call and this PR gates it.

⚠️ **Operational note for CI consumers:** because this now *correctly* fails loud, a `specsync diff` step running in a **shallow checkout** (the GitHub Actions default) will start failing where it previously passed silently — the auto-detected base `origin/<GITHUB_BASE_REF>` isn't present in a shallow clone. Fix: `actions/checkout` with `fetch-depth: 0`. The error message now says this.

**Follow-ups (pre-existing, out of scope — not blocking):**
- Export-delta path treats an unparseable/missing source file as "no exports", so documented symbols get falsely listed as "removed". This is a *parser*-layer fail-open (no git involved) and is **informational-only** — `diff` exits 0 regardless of drift, so it doesn't gate CI like the bad-base case this PR fixes.
- git<2.24 lacks `--end-of-options` (a pre-existing token); realistically no impact on modern toolchains.

Also corrected spec invariant #3 (it referenced a `git show` that doesn't exist).

PR #316, opened 2026-07-04
Fix: reject path-traversal module names in `add-spec`/`scaffold`

0xLeif:
Medium-tier audit finding (#14) — a **path traversal** in spec scaffolding.

## Problem

`cmd_add_spec` and `cmd_scaffold` wrote the user-supplied `module_name` **verbatim** into paths — `<specs_dir>/<name>/<name>.spec.md`, and joined onto source dirs — with no validation. A name containing `../` or an absolute path escaped the project root:

```
$ specsync add-spec "../../PWNED/evil"
  ✓ Created specs/../../PWNED/evil/../../PWNED/evil.spec.md
EXIT=101                                  # wrote files OUTSIDE the project, then panicked
```

The spec file **and** its companion directory landed outside the project root. Path traversal writing files to arbitrary locations.

## Fix

Added `validate_module_name`: a module name must be a **single path segment**. Empty, path separators (`/`, `\`), `.`/`..`, and absolute paths are refused with a clear error and `exit 1`, **before any filesystem write**. Both entry points are gated. Legitimate names (`auth`, `auth-service`, `user_profile`) are unaffected.

```
$ specsync add-spec "../../PWNED/evil"
invalid module name `../../PWNED/evil`: use a single name without path separators (`/`, `\`), `.`/`..`, or an absolute path
EXIT=1                                    # nothing created outside the project
```

## Tests

- Unit: `validate_module_name` accepts plain names, rejects separators / `..` / absolute / empty.
- Integration: `scaffold_rejects_module_name_path_traversal` — both `add-spec` and `scaffold` with `../../escape/evil` fail loud with "invalid module name" and write nothing outside the project root.
- Documented the new invariant (#5) + Error Cases in `cmd_scaffold.spec.md`, updated testing.md.
- **734 unit + 170 integration**, `cargo fmt --check` clean, `cargo clippy --bin specsync -- -D warnings` clean, self-check 60/60 at 100%.

### DIFF
--- a/specs/cmd_new/cmd_new.spec.md
+++ b/specs/cmd_new/cmd_new.spec.md
@@ -55,6 +55,7 @@ Implements the `specsync new` command. Quick-creates a minimal spec with auto-de
 | Spec already exists | Exits 1 |
 | No source files found | Creates spec with empty `files:` and prints a ⚠ explaining that the `files:` list must be filled in before `check` passes |
 | Dir creation fails | Exits 1 |
+| Invalid module name (path separator, `.`/`..`, absolute/drive-relative, control chars) | Refused via `validate_module_name` before any write; prints `invalid module name …` and exits 1 (no path traversal) |
 
 ## Dependencies
 
--- a/specs/cmd_scaffold/cmd_scaffold.spec.md
+++ b/specs/cmd_scaffold/cmd_scaffold.spec.md
@@ -34,6 +34,7 @@ Implements `specsync add-spec` and `specsync scaffold` commands. Creates new spe
 2. `cmd_scaffold` supports custom templates and auto-appends to registry
 3. Neither overwrites existing specs
 4. Companion files (tasks.md, context.md, requirements.md, testing.md) are always generated with guided starter content; design.md is generated only when `companions.design` is enabled in config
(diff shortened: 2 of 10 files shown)


0xLeif:
**Re-review after fixes: READY, 90%, zero blocking.** Both prior blocking issues confirmed fixed:

1. ✅ **cmd_new** — now calls `validate_module_name` as its first statement (before `load_config`/any write). `specsync new "../../PWNED" --full` now exits 1 with nothing written outside.
2. ✅ **Windows drive-relative `C:foo`** — the validator dropped `is_absolute()` for a single-`Component::Normal` requirement, which is platform-aware and rejects the `Prefix` component on Windows (covered by a `#[cfg(windows)]` test).

All four scaffolding entry points (`new`, `add-spec`, `scaffold`, `wizard`) route through the shared `validate_module_name`; no legit single-segment name regresses (hyphen/underscore/dotted/unicode verified). Exhaustive `.join()`/`create_dir_all`/`fs::write` sweep found no other user-name escape.

**Accepted follow-ups (non-blocking, no action needed here):**
- `import` derives the module name via `slugify` → `[a-z0-9-]`, so it's safe by construction; routing it through the shared validator would be optional defense-in-depth.
- `scaffold --dir` intentionally lets the user pick an output directory (trusted input by design) — not a name-traversal.

PR #317, opened 2026-07-04
Fix: `**/**` exclude pattern panics the coverage glob matcher

0xLeif:
Medium-tier audit finding (#7) — a **panic** (crash) in the coverage exclude-pattern matcher.

## Problem

The `**/dir/**` branch of the exclude matcher extracts the middle via `&pattern[3..pattern.len()-3]`. For the degenerate `**/**` (len 5), the `**/` prefix and `/**` suffix overlap, so the range reverses to `[3..2]` and Rust panics:

```
$ specsync check    # with exclude_patterns = ["**/**"]
thread 'main' panicked at src/validator.rs:1364: byte range starts at 3 but ends at 2
```

## Fix

Extract the middle by peeling both ends — `strip_prefix("**/")` then `strip_suffix("/**")`, falling back to `""`. For `**/**` this yields an empty `dir_part`, and `rel_str.contains("")` is always true, so `**/**` matches every path (the correct "match everything" meaning) — no panic. Normal patterns (`**/gen/**` → `gen`) are byte-for-byte unchanged.

## Reproduced

`check` with `exclude_patterns = ["**/**"]` went from a panic (exit 1) to cleanly excluding all source files (`File coverage: 0/0`).

## Tests

- `test_compute_coverage_double_star_exclude_no_panic`: `**/**` no longer panics and excludes all files; `**/gen/**` still excludes only that directory.
- Documented the supported exclude-glob forms in the validator spec (invariant #4).
- **735 unit + 170 integration**, `cargo fmt --check` clean, `cargo clippy --bin specsync -- -D warnings` clean, self-check 60/60 at 100%.

### DIFF
--- a/specs/validator/validator.spec.md
+++ b/specs/validator/validator.spec.md
@@ -39,7 +39,7 @@ Core validation engine for spec-sync. Validates individual spec files against so
 1. Validation is bidirectional: spec documenting non-existent exports = ERROR; code exports not in spec = WARNING
 2. Missing frontmatter fields (module, version, status, files) are errors, not warnings
 3. Cross-project refs (`owner/repo@module`) are skipped during local validation — only checked by `specsync resolve`
-4. Coverage computation excludes test files and configured exclude patterns
+4. Coverage computation excludes test files and configured exclude patterns. Exclude globs support `**/dir/**` (path contains `dir`), `**/*.ext` (suffix), and `**/name` (filename); a degenerate `**/**` matches every path (empty middle) and is handled without panicking
 5. Source file discovery respects `source_extensions` config — empty means all supported languages
 6. `find_spec_files` returns sorted results
 7. Schema table extraction supports configurable regex patterns via `schema_pattern` config
--- a/src/validator.rs
+++ b/src/validator.rs
@@ -838,6 +838,38 @@ mod tests {
     use super::*;
 
     #[test]
+    fn test_compute_coverage_double_star_exclude_no_panic() {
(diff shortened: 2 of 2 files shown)


0xLeif:
**Adversarial review: READY, 95%, zero blocking, no new defects.**

- **Behavior-preserving:** verified byte-identical `dir_part` vs the old slice for all normal patterns (both `**/` and `/**` are 3 ASCII bytes, peeled on char boundaries — unicode dirs like `**/日本/**` fine). Only `**/**` diverges (old: panic; new: `dir_part=""` → matches all). `**//**` already yielded `""` pre-fix, so `**/**` isn't uniquely "exclude-all".
- **No other panics:** probed 18 inputs (`**/`, `/**`, `**`, `**/**/**`, `**//**`, `**/*.ts`, empty, unicode) via real `coverage` runs — zero panics, exclusion outcomes match pre-fix semantics.
- Regression test + full `validator::` suite pass; clean build.

**Pre-existing follow-ups (out of scope for this panic fix):**
- **Vacuous 100% coverage:** when a config yields 0 source files (via `**/**`, `**/*.ext` excluding everything, or a wrong `source_dirs`), `coverage_percent` is 100 (`validator.rs:1527`) so `--require-coverage 100` passes vacuously. A "fail-loud" hardening would warn on "0 source files under an explicit `--require-coverage`" — filed as its own follow-up, not scoped here.
- Exclude matcher uses loose **substring** matching (`**/a/**` excludes any path containing `a`), unchanged by this PR.

PR #318, opened 2026-07-04
Fix: `--require-coverage` fails loud on zero source files (no vacuous pass)

0xLeif:
Follow-up surfaced by the #317 review — a **fail-open** in the coverage gate.

## Problem

When 0 source files are discovered, coverage is reported as a vacuous **100%** (`validator.rs`: `if all_source_files.is_empty() { 100 }`). So `--require-coverage N` was silently satisfied. A misconfigured project — a wrong `source_dirs`, or an over-broad `exclude_patterns` like `**/**` — would **pass a coverage gate in CI while measuring nothing**.

## Fix

`compute_exit_code` / `exit_with_status` now fail loud (exit 1) when `--require-coverage N` is set with `N > 0` **and** 0 source files were found, printing a message that points at `source_dirs` / `exclude_patterns`:

```
--require-coverage 100%: no source files were found to measure coverage against —
check `source_dirs` and `exclude_patterns` (an empty or over-broad config reports
coverage as a vacuous 100%)
```

A require of `0`, or no `--require-coverage` flag, still passes (nothing is demanded). Real projects at 100% (non-zero files) are unaffected — verified this repo's own `check --require-coverage 100` still passes.

## Test fixture correction

`strict_mode_exits_0_when_coverage_meets_threshold` (and the sibling require-coverage unit tests) built a **contradictory** report — 0 source files but 85% — via `empty_coverage()`. They now use a realistic `coverage_pct()` helper so they exercise the *percentage* gate, not the new zero-sources gate.

## Tests

- Unit: `require_coverage_fails_loud_on_zero_source_files` (0 files + `Some(100)` → 1; `Some(0)` → 0; `None` → 0).
- Integration: a `**/**`-excluded project fails `check --require-coverage 100` with "no source files were found".
- **736 unit + 171 integration**, `cargo fmt --check` clean, `cargo clippy --bin specsync -- -D warnings` clean, self-check 60/60 at 100%.

### DIFF
--- a/specs/commands/commands.spec.md
+++ b/specs/commands/commands.spec.md
@@ -79,7 +79,7 @@ Shared command infrastructure used by all CLI subcommands. Provides config loadi
 4. In text mode, draft specs show explicit "Section validation skipped (status: draft)" and "Export validation skipped (status: draft)" notices instead of misleading green checkmarks, plus a closing hint to set `status: active`
 5. Failing checks render negated labels (e.g. "✗ Frontmatter invalid"), never a ✗ next to a passing label
 4. Exit code logic by enforcement mode: Warn → always 0; EnforceNew → 1 if unspecced files; Strict → 1 on errors, also 1 on warnings when `--strict`
-5. `--require-coverage N` triggers exit 1 if file coverage percent < N regardless of enforcement mode
+5. `--require-coverage N` triggers exit 1 if file coverage percent < N regardless of enforcement mode. When `N > 0` but 0 source files were discovered (empty/misconfigured `source_dirs` or an over-broad `exclude_patterns`), the gate fails loud (exit 1) rather than passing on the vacuous 100% reported for an empty source tree
 6. `create_drift_issues` groups errors by spec path and creates one GitHub issue per spec, not per error
 
 ## Behavioral Examples
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -604,10 +604,17 @@ pub fn compute_exit_code(
             }
         }
     }
-    if let Some(req) = require_coverage
-        && coverage.coverage_percent < req
-    {
-        return 1;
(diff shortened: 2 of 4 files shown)


0xLeif:
**Adversarial review: READY_WITH_NITS, 93%, zero blocking, no real defects.**

- **No false positives:** real projects at 100% + `--require-coverage 100` → exit 0 (incl. this repo's self-check, 76/76). Percentage-gate vs zero-source-gate branches select correctly (`85 > 80` on real files → percentage message; 0 files → zero-source message). `--require-coverage 0` and no-flag both exit 0.
- **Docs-only repo (fail-loud judgment):** acceptable, not over-aggressive — `--require-coverage N` is an opt-in "I have source that must be specced" assertion; 0 discovered sources means a broken `source_dirs`/`exclude_patterns` or a meaningless requirement, both worth surfacing, with `--require-coverage 0` as the escape hatch.
- **Completeness:** the only two `coverage_percent < req` gate sites (both in `mod.rs`) are now guarded. `loc_coverage_percent` never gates (display/JSON only). `diff`/`mcp` have no coverage gate. `check`/`comment`/`coverage`/`generate`/`score` all route through the two fixed functions; `watch` shells out to `check` and inherits the fix. No remaining vacuous path.
- **Fixtures:** `coverage_pct()` correctly isolates the percentage gate; the switch from `empty_coverage()` was necessary.

**Cosmetic follow-up (non-blocking, pre-existing):** on the no-specs (`check.rs:141`) and warm-cache (`check.rs:225`) paths, the gate uses the silent `compute_exit_code`, so a zero-source failure exits 1 correctly but without the `no source files were found…` explanation (only the normal `exit_with_status` path prints it). The user sees `0/0 (100%)` next to a bare exit 1. Exit code is correct on all paths; only the message is thinner on two. Candidate for a one-line follow-up.

PR #319, opened 2026-07-04
Fix: `is_test_file` misclassifies sources under a test-named ancestor

0xLeif:
Medium-tier audit finding (#11) — a **silent wrong result** in coverage.

## Problem

`is_test_file` walked **all** components of the path to decide whether a file lives in a test directory. Callers pass **absolute** paths (coverage and generator walk `root.join(src_dir)`, and `root` is canonicalized), so a project living under a directory whose name is in `TEST_DIR_NAMES` — `test`, `tests`, `spec`, `specs`, `testing`, `__tests__` — had **every source file classified as a test and silently excluded**.

A project at `~/spec/myproject` (or a repo checked out under a `test/` folder, common in CI) then reported:

```
File coverage: 0/0 (100%)     # real source silently excluded
```

That's a silent wrong result — and combined with the new `--require-coverage` zero-source gate, it fails loud on the (wrongly) empty source set.

## Fix

`is_test_file` now takes `root` and bounds the directory check to components **below** `root` (via `strip_prefix`). Ancestors above the project are ignored; filename-pattern detection is unchanged; in-project test dirs (`src/tests/`, `__tests__/`) are still excluded. Threaded `root` through `find_source_files` (validator) and `find_module_source_files` (generator) and their call sites (6 call sites total).

## Reproduced

A project under a `spec/` parent went from `0/0 (100%)` → `1/1 (100%)`; an in-project `src/tests/helper.ts` is still excluded.

## Tests

- `is_test_file` unit tests: ancestor above root ignored (`/home/user/spec/proj/src/app.ts`, `/ci/tests/checkout/repo/src/lib.rs`); in-project `tests/`/`__tests__/` + `.test.ts`/`.spec.ts` filenames detected; plain source not a test.
- Updated the exports spec signature + added a scenario.
- **740 unit + 171 integration**, `cargo fmt --check` clean, `cargo clippy --bin specsync -- -D warnings` clean, self-check 60/60 at 100%.

### DIFF
--- a/specs/exports/exports.spec.md
+++ b/specs/exports/exports.spec.md
@@ -40,7 +40,7 @@ Language-aware export extraction from source files. Auto-detects the programming
 |----------|-----------|---------|-------------|
 | `get_exported_symbols` | `file_path: &Path` | `Vec<String>` | Extract exported symbol names from a source file, auto-detecting language from extension |
 | `get_exported_symbols_with_level` | `file_path: &Path, level: ExportLevel` | `Vec<String>` | Extract exports with configurable granularity — Type (declarations only) or Member (all symbols) |
-| `is_test_file` | `file_path: &Path` | `bool` | Check if a file is a test file based on language-specific naming conventions |
+| `is_test_file` | `file_path: &Path, root: &Path` | `bool` | Check if a file is a test file by filename convention or a test directory *within `root`*; the directory check is bounded to project-relative components so ancestors above `root` (e.g. a parent dir named `spec`/`test`) do not misclassify ordinary sources |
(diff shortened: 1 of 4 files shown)


0xLeif:
**Adversarial review: READY, 95%, no defects.** Core fix verified end-to-end (project under a `spec/` ancestor: `0/0` → `1/1`); the strip_prefix regression risk is ruled out — an in-project `src/tests/` file is still excluded (proving `strip_prefix(root)` succeeds and the dir check runs), and a normal project's coverage is byte-identical to main (76/76). Symlinks unaffected (WalkDir yields root-prefixed logical paths). All 5 call sites pass the true project root; no other ancestor-walking test-detection path exists. 911 tests pass; self-check green.

PR #321, opened 2026-07-04
Fix: parse TOML literal (single-quoted) strings in config

0xLeif:
Medium-tier audit finding (#12) — a **silent wrong result** in config parsing.

## Problem

The hand-rolled config reader only recognized double-quoted `"..."` strings; it never tracked `'...'`, which is **equally valid TOML**. Any value written with single quotes was mis-parsed and the setting silently dropped.

Verified: a `.specsync.toml` with
```toml
specs_dir = 'specs'
source_dirs = ['lib']
```
made `specsync coverage` report **`File coverage: 0/0 (100%)`** on a project with real source — the parser kept the quotes and scanned a directory literally named `'lib'`, found nothing, and reported a vacuous 100% (same class as #305 / #310 / #318). A `]` inside a single-quoted element (`['a]b']`) also closed the array early, since the close-scan didn't treat `'` as a string delimiter.

## Fix

The four string-scanning helpers now track **both** TOML string kinds — basic `"..."` (backslash-escaped) and literal `'...'` (no escapes, verbatim):

- `strip_inline_comment` — a `#` inside `'...'` is content, not a comment.
- `find_toml_array_close` — a `[`/`]` inside `'...'` doesn't count toward depth.
- `split_toml_array_items` — a `,` inside `'...'` doesn't split an element.
- `parse_toml_string` — `'...'` is taken verbatim (TOML literal strings do no escape processing), so a Windows path or regex keeps its backslashes; `"..."` still unescapes exactly as before.

Every scalar and array value routes through these helpers, so this covers the whole config surface in one place. Multi-line `'''`/`"""` strings remain out of scope (unchanged).

## Reproduced

- `source_dirs = ['lib']` → `0/0 (100%)` **→ `1/1 (100%)`**.
- A literal-quoted exclude `'**/gen/**'` now applies (gen excluded).
- Mixed quotes (`["src", 'gen']`) and a `]`/`,` inside a literal element parse intact.

## Tests

- 8 new `config` tests: single-quoted scalar/array, mixed quotes, literal glob containing `]`, `,`/`]` inside literals stay intact, `#` inside a literal not a comment, backslash preservation, and an end-to-end single-quoted `.specsync.toml`.
- Added config spec invariant #8 (both string kinds for scalar + array values).
- **748 unit + 171 integration**, `cargo fmt --check` clean, `cargo clippy --bin specsync -- -D warnings` clean, self-check 60/60 at 100%.

### DIFF
--- a/specs/config/config.spec.md
+++ b/specs/config/config.spec.md
@@ -59,6 +59,7 @@ The configuration file supports the following top-level sections:
 5. `detect_source_dirs` falls back to `["src"]` if no source files are found
 6. Root-level source files (no subdirectories) produce `["."]` as source dirs
 7. TOML parsing is zero-dependency — uses line-by-line string parsing, not a TOML library
+8. The reader accepts both TOML string kinds for scalar and array values: basic `"..."` strings (backslash escapes decoded) and literal `'...'` strings (taken verbatim, no escape processing); a `#`, `,`, `[`, or `]` appearing inside either kind is treated as content, not as a comment or array structure
 
 ## Behavioral Examples
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -941,19 +941,30 @@ fn load_toml_config(config_path: &Path, root: &Path) -> SpecSyncConfig {
 ///   `120 # seconds`     →  `120`
 ///   `"has # inside"`    →  `"has # inside"`
 fn strip_inline_comment(s: &str) -> String {
-    let mut in_string = false;
+    // Track both TOML string kinds: basic `"..."` (backslash-escaped) and literal
+    // `'...'` (no escapes). A `#` inside either is content, not a comment.
+    let mut in_basic = false;
+    let mut in_literal = false;
     let mut prev_backslash = false;
     for (i, c) in s.char_indices() {
-        if in_string {
+        if in_basic {
             if c == '"' && !prev_backslash {
-                in_string = false;
+                in_basic = false;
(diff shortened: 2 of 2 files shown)


0xLeif:
**Adversarial review: READY_WITH_NITS, confidence 88 → resolved.**

Confirmed on the built branch: `"..."` handling byte-identical to main (no regression), core fix works end-to-end (`source_dirs = ['lib']`: 0/0 → 1/1; literal exclude `'**/gen/**'` applies; mixed quotes parse), state machine correct (a `"` inside a literal / `'` inside a basic are content; literal strings don't unescape; `prev_backslash` resets), no panics/hangs on malformed input, 171 tests + self-check green.

**Nit fixed (`eeb8e79`):** the review caught an incomplete-fix gap — the multi-line-array continuation guard (from #305) only accepted `"`/`]`-led continuation lines, so a formatter-style multi-line array with single-quoted items collapsed to `["["]` (the same vacuous-coverage failure this PR targets). Extended the guard to accept `'`-led lines; added `test_load_config_toml_multiline_single_quoted_array`. Verified 0/0 → 1/1.

**Accepted edge (not fixed):** an apostrophe in an *unquoted* value with a trailing inline comment can swallow the comment — but this is exactly symmetric with the long-standing `"` behavior on main and is avoided by quoting the value. Garbage-in on invalid TOML degrades gracefully (no panic).

PR #323, opened 2026-07-04
Fix: gate JSON/visualization paths that bypassed --strict/--enforcement/--require-coverage

0xLeif:
Audit **Batch 1** — three "a gate must apply on *every* path" bugs, same class as the already-fixed coverage M1 / score JSON bypass.

## Fixed

- **[high] `generate --format json` bypassed the gates.** Both JSON early-returns (`cmd_generate_all` + `cmd_generate_batch`) exited solely on AI-generation success and never fed coverage/validation/`--strict`/`--enforcement`/`--require-coverage` into the exit code (`compute_exit_code` wasn't even imported). A CI step using `generate --format json` as a gate green-lit exactly the broken states it exists to catch. Both branches now recompute post-generation coverage + validation and gate via `compute_exit_code` (prints nothing → stdout stays valid JSON), matching the text path.
- **[medium] `generate --format json` emitted a non-JSON line.** The "No existing specs found…" diagnostic printed to stdout even in JSON mode, prepending junk that broke any parser. Suppressed under `--format json`.
- **[low] `deps --strict --mermaid`/`--dot` skipped the strict gate.** The visualization branch early-returned before the gate, so an undeclared import that fails `deps --strict` silently passed. The diagram still prints; the gate (errors always, warnings under `--strict`) now applies, note on stderr.

## Reproduced (each exit 0 → exit 1, JSON preserved)

- empty source dir → `generate --require-coverage 50 --format json` now fails (vacuous coverage)
- spec with a missing-file error → `generate --enforcement strict --format json` (+ `--batch`) now fails
- undeclared import → `deps --strict --mermaid` now fails, diagram still on stdout

## Tests

4 integration tests + **749 unit / 175 integration**, `cargo fmt --check` clean, `clippy --bin` clean, self-check 60/60 at 100%.

Part of the post-v4.7.1 audit sweep (31 findings across ~10 batches).

### DIFF
--- a/src/commands/deps.rs
+++ b/src/commands/deps.rs
@@ -24,6 +24,23 @@ pub fn cmd_deps(root: &Path, strict: bool, format: types::OutputFormat, mermaid:
                 "Hint:".yellow()
             );
         }
+        // A visualization request must still honor the gate flags: without this,
+        // `deps --strict --mermaid`/`--dot` silently exited 0 on the same undeclared
+        // imports / cycles / missing deps that the non-visual path fails on. The
+        // diagram remains the only thing on stdout; the gate note and exit go to
+        // stderr / the exit code, consistent with the normal `deps` path.
+        let report = deps::validate_deps(root, &config.specs_dir);
+        let strict_fail = strict && !report.warnings.is_empty();
+        if strict_fail {
+            eprintln!(
+                "{}: {} dependency warning(s) treated as errors",
+                "--strict mode".red(),
+                report.warnings.len()
+            );
+        }
+        if !report.errors.is_empty() || strict_fail {
+            process::exit(1);
+        }
         return;
     }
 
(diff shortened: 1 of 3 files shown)


0xLeif:
**Adversarial review: READY_WITH_NITS, 92, no blocking defects.** Verified end-to-end: all three fixes work (generate json/text gate identically on require-coverage + enforcement; clean JSON with no specs; deps --strict --mermaid/--dot gate with diagram still on stdout). Critical risk cleared — the post-generation recompute means `generate --require-coverage 100 --format json` still exits 0 when generate scaffolds the missing spec (gates on POST-gen state, text and JSON agree). deps --mermaid exiting 1 on a genuine error (cycle/missing-dep) matches base `deps` behavior (intended/consistent). Nits (non-blocking): JSON gate-fail body carries no reason (exit code is the signal; overlaps Batch 9), recompute block is a DRY candidate, viz path builds the graph twice.

PR #324, opened 2026-07-04
Fix: export-extraction fail-open (diff + score treated unreadable files as export-free)

0xLeif:
Audit **Batch 2** — one root cause + two gating callers.

## Root cause

`get_exported_symbols_full` returned an empty `Vec` for **both** an unrecognized extension **and** a read failure (missing / permission / non-UTF-8) — indistinguishable from a file that genuinely exports nothing. Callers that gate on the result silently treated an *unanalyzable* file as having no API.

## Fixed

- **[medium] `diff` fail-open.** A changed source file whose exports couldn't be read contributed zero exports → real new API dropped → "no drift", exit 0. This defeated the command and its own spec invariant ("a failed comparison must not silently pass in CI"), previously implemented only for the git-diff exit path.
- **[medium] `score` fail-open.** The API dimension scored an empty extraction as a **perfect** "no exports to document" (20/20), so a missing/non-UTF-8 `files:` entry inflated the total — which feeds the lifecycle `min_score` gate.

## Approach

Added `ExportScan { Parsed(Vec<String>), UnknownLanguage, Unreadable }` + `scan_exported_symbols[_full]`. `get_exported_symbols_full` is now a thin wrapper that still coerces to an empty vec, so **all existing callers are unchanged** (incl. `check`, which already reports unreadable files loudly). `diff` records `Unreadable` entries as inconclusive, surfaces them, and fails loud (exit 1); `score` withholds API credit (0/20 + "could not analyze") when empty *only* because a file couldn't be read. `UnknownLanguage` (a `.md`/`.sql` legitimately listed) stays benign in both.

## Reproduced

- `diff` on a non-UTF-8 changed file: silent "no drift"/exit 0 → **inconclusive/exit 1**.
- `score` on a missing `files:` entry: **API 20/20 (total ~95) → API 0/20 (total 41)**. A readable export-free file still scores 20/20.

## Tests

`scan` unit test (Parsed vs UnknownLanguage vs Unreadable incl. non-UTF-8); diff fail-loud + score withhold-credit integration tests. Documented `ExportScan` + `scan_*` in the exports spec. **750 unit + 177 integration**, fmt / clippy / self-check 100%.

Part of the post-v4.7.1 audit sweep (Batch 2 of ~10).

### DIFF
--- a/specs/cmd_diff/cmd_diff.spec.md
+++ b/specs/cmd_diff/cmd_diff.spec.md
@@ -36,6 +36,7 @@ Implements the `specsync diff` command — shows which specs are affected by sou
 4. Changed files not covered by any spec are listed separately
 5. Text output delegates to `output::print_diff_markdown`
 6. A non-zero `git diff` exit (bad base ref, not a git repository) fails loud with exit code 1 — it is never treated as an empty change set ("No files changed"), so a failed comparison cannot silently pass in CI
+7. A spec whose `files:` includes a source file that exists-or-should but cannot be read (missing, or not valid UTF-8) is inconclusive: the file is surfaced (`inconclusive_files`), its unreliable export deltas are suppressed, and `diff` exits 1 — an unreadable source must not be silently treated as export-free "no drift" (a non-source file, e.g. a `.md`/`.sql` listed in `files:`, is not a failure)
 
 ## Behavioral Examples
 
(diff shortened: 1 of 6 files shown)


0xLeif:
**Adversarial review: READY_WITH_NITS, 92 — no correctness defects, no backward-compat regressions.** Verified line-for-line backward compat (the 4 existing callers of get_exported_symbols[_full] are unchanged; wrapper is byte-identical) and all core fixes end-to-end (diff inconclusive/exit-1 on non-UTF-8 + missing; no false positive on normal or .md/.sql files; score API 20/20→0/20 on unreadable, benign cases preserved). Applied both recommended nits in the follow-up commit: (1) suppress the unreliable new/removed_exports for inconclusive entries (was a bogus 'removed apiFn' false-deletion in JSON), (2) documented the exit-1 trigger as cmd_diff spec invariant #7.

PR #325, opened 2026-07-04
Fix: file-read fail-loud for config + deps (unreadable-but-present files silently degraded)

0xLeif:
Audit **Batch 3** (config & deps half) — three defects of the same shape: *a file that **exists** but can't be read as UTF-8 was silently degraded instead of failing loud.*

## Fixed

- **[medium] Unreadable config → silent defaults.** `load_toml_config` / `load_json_config` mapped a read failure to `SpecSyncConfig::default()` with no signal, **downgrading enforcement** (strict→warn, exit 1→0) on a config that was present but not valid UTF-8. Callers reach these only after an `.exists()` check, so a `None` on a still-existing file now warns loudly (naming the file) before falling back. Same for the optional local override (`config.local.toml`). An *absent* config stays silent (expected).
- **[low] `deps` drops an unreadable source's imports.** `check_undeclared_imports` did `Err(_) => continue`, so a non-UTF-8 declared source contributed no imports and `deps --strict` could pass while hiding undeclared-import violations. Now a hard error (mirrors `validator.rs`'s source-read policy); `cmd_deps` exits 1.
- **[medium] `deps` drops an unreadable spec from the graph.** `build_dep_graph` did `Err(_) => continue`, silently removing the node — defeating cycle and missing-dependency detection for that module. Refactored into an internal `build_dep_graph_checked` that also returns read-failure messages; `validate_deps` extends `report.errors`. `build_dep_graph` keeps its signature for the non-gating visualization / topological-order callers.

## Reproduced

- non-UTF-8 config → loud `exists but could not be read` warning (was silent);
- non-UTF-8 declared source → `deps` error + **exit 1** (was exit 0);
- non-UTF-8 spec → `deps` error + **exit 1** (was exit 0, node silently dropped).

## Tests

3 integration regressions + config spec invariant #9 + deps spec invariant #7. **750 unit + 180 integration**, fmt / clippy / self-check 100%.

## Scope note

The **schema-discovery fail-open** (`build_schema` / `get_schema_table_names` silently skipping unreadable migration files, weakening DB-column validation) is the same theme but needs signature changes across `schema.rs` + `validator.rs` + the check command — it ships as its own follow-up PR to keep this one focused.

Part of the post-v4.7.1 audit sweep (Batch 3).

### DIFF
--- a/specs/config/config.spec.md
+++ b/specs/config/config.spec.md
@@ -1,6 +1,6 @@
 ---
 module: config
-version: 1
+version: 2
 status: stable
 files:
   - src/config.rs
@@ -86,7 +86,8 @@ The configuration file supports the following top-level sections:
 
 | Condition | Behavior |
 |-----------|----------|
-| Config file unreadable | Falls back to `SpecSyncConfig::default()` |
+| Config file exists but unreadable (e.g. not valid UTF-8) | Prints a warning naming the file to stderr, then falls back to `SpecSyncConfig::default()` (fail-loud, not silent) |
+| Config file absent | Silently uses `SpecSyncConfig::default()` with auto-detected source dirs (expected) |
 | Malformed JSON config | Prints warning to stderr, falls back to defaults |
 | Empty project root | Returns `["src"]` as source dirs |
 
@@ -117,3 +118,4 @@ The configuration file supports the following top-level sections:
 | 2026-03-28 | Document discover_manifest_modules |
 | 2026-04-06 | Document github config section, rules section, and full config file structure |
 | 2026-07-03 | Document `read_config_file` (leading-BOM-tolerant config read, shared with `migrate`) |
+| 2026-07-06 | Document fail-loud behavior for present-but-unreadable config (invariant 9): unreadable config/local-override files now warn loudly before falling back to defaults; clarified Error Cases table |
--- a/specs/config/config.spec.md
+++ b/specs/config/config.spec.md
@@ -60,6 +60,7 @@ The configuration file supports the following top-level sections:
 6. Root-level source files (no subdirectories) produce `["."]` as source dirs
 7. TOML parsing is zero-dependency — uses line-by-line string parsing, not a TOML library
 8. The reader accepts both TOML string kinds for scalar and array values: basic `"..."` strings (backslash escapes decoded) and literal `'...'` strings (taken verbatim, no escape processing); a `#`, `,`, `[`, or `]` appearing inside either kind is treated as content, not as a comment or array structure
+9. A config file that is absent is expected — defaults apply silently. But a config file that **exists yet cannot be read** (e.g. not valid UTF-8) fails loud: a warning naming the file is printed and built-in defaults are used, rather than silently reverting to defaults (which would downgrade enforcement — strict→warn, exit 1→0 — with no signal). The same applies to the optional local override file (`config.local.toml`)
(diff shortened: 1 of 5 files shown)


0xLeif:
**Adversarial review: READY_WITH_NITS, 93 — no defects found.**

- **False-positive check (the critical one): CLEAN.** No config → zero warnings; normal readable config (all 4 file forms + local override) → zero warnings; only a present-but-unreadable config warns. Every loader call site is `.exists()`-guarded before reaching the new warning gate (load_config precedence chain + both migrate.rs load_config_from_path sites), so absent stays silent and present-unreadable warns — verified structurally and empirically.
- **Security: CLEAN.** git diff touches zero `ai_command` lines; the on-disk ai_command stripping + local-merge refusal are byte-for-byte untouched.
- **Deps: CLEAN.** non-UTF-8 source → exit 1 (no false positive on readable); non-UTF-8 spec → exit 1 in text+JSON; --strict and --strict --mermaid gate.
- **Backward compat: CLEAN.** `deps` text/mermaid/dot/json output byte-identical between main and branch; validate_deps calls build_dep_graph_checked once (no double I/O); non-gating callers unaffected.
- 750 unit + 180 integration pass; self-check 100%; build_dep_graph_checked is private (no new public API).

**Non-blocking nit (point 6):** the config warning is stderr-only and CI gates on exit code, so an unreadable *enforcement* config downgrades to defaults and still exits 0. This is consistent with the pre-existing JSON-parse-error branch (also warn+default), so it ships as-is; escalating an unreadable config to a hard non-zero exit for gating commands is noted as a candidate future hardening.

Kyntrin (changes requested):
Thanks for the focused regression coverage here. The code change looks directionally right and CI is green, but the spec updates need to be completed before merge.

Requested changes:

1. `specs/deps/deps.spec.md` now has a new invariant saying unreadable declared source/spec files are hard errors, but the Error Cases table still says `Source file unreadable | Skipped during import extraction`. That contradicts the new behavior in `validate_deps` / `cmd_deps`; please update the Error Cases row to match the hard-error behavior, and add an unreadable-spec row if you want the table to cover both new cases.

2. Both changed specs add new behavioral contract, but neither spec was version-bumped or given a Change Log entry. Per this repo's AGENTS.md workflow, after spec changes we should increment the spec `version` field and add a dated Change Log entry. Please update `specs/config/config.spec.md` and `specs/deps/deps.spec.md` accordingly.

Non-blocking note: `specs/config/config.spec.md` could also make the Error Cases table more explicit that unreadable config files now warn loudly before falling back to defaults. I’m not blocking on the existing fallback wording as long as the version/changelog update lands, but tightening it would prevent future ambiguity.

Validation I ran locally: `cargo run --quiet -- check --strict` passed with 60 specs checked, 60 passed.

0xLeif:
@Kyntrin Addressed in 6e8c28c:
1. `specs/deps/deps.spec.md` Error Cases table updated — the `Source file unreadable` row now says hard error (recorded in `report.errors`, `cmd_deps` exits 1), and a new unreadable-spec row was added.
2. Version bumps + Change Log entries landed on both specs (`deps` 2→3, `config` 1→2).
3. Also tightened the config Error Cases table to state that a present-but-unreadable config warns loudly before falling back (plus an explicit absent-config row).

`specsync check --strict` passes 60/60.

PR #327, opened 2026-07-04
Record attest provenance on merged commits

0xLeif:
From the trust-tooling audit: spec-sync never ran attest, so its /trust/ attest pip was dark. Adds a dedicated **main-only** `attest` job (after test, fmt, audit, spec-check) that records a provenance attestation and pushes `refs/notes/attest`. Public Linux binary, checksum-verified, no token; additive and best-effort.

https://claude.ai/code/session_01ApSS1xj22zGwbUSkRs3mmE

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,7 +242,7 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       # After the gates pass, record a provenance attestation on this commit and
       # push it to refs/notes/attest, so spec-sync builds a trust ledger (and its
       # /trust/ attest pip lights). attest ships a public Linux binary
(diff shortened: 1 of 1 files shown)


Kyntrin (changes requested):
The workflow addition is small and the PR checks are green, but I’m requesting one change before merge.

Requested change:

The new `attest` job records a `proceed` attestation after only `test`, `fmt`, `audit`, and `spec-check` complete. Existing CI also has `validate-action`, `coverage`, `site`, and `vscode-extension` jobs. On a push to `main`, any of those omitted jobs could fail while the attestation job still runs and pushes a successful provenance note for the commit. That makes the ledger stronger than the actual CI state.

Please include the full set of merge-gating CI jobs in `needs` before recording the attestation, or make the attestation note/verdict explicitly narrower so it cannot be read as “all CI passed.”

Non-blocking notes: the job’s best-effort behavior is implemented as advertised, and it correctly skips on PRs. Consider using `actions/checkout@v5` for consistency with the rest of the workflow, but that is not a blocker.

0xLeif:
@Kyntrin Addressed:
- The `attest` job's `needs` now gates on the full merge-gating set — `test, fmt, audit, coverage, site, vscode-extension, validate-action, spec-check` — so a `proceed` note can no longer be recorded when any real check failed (done in 4fab4f4).
- Follow-up 0288b34: updated the now-stale `--note` text (it still enumerated only `test + fmt + audit + spec-check`) to list the full gate set so the ledger entry can't be mis-read, and bumped the job's `actions/checkout@v4`→`v5` for consistency (your non-blocking note).

PR #331, opened 2026-07-10
Add: 33-language export extraction + Rust pub(crate) filtering + TypeScript export = support

0xLeif:
## Summary
- Expand regex export backends from 12 to 33 languages (C, C++, Scala, Crystal, Nim, Erlang, Elixir, Perl, Lisp, Haskell, Lua, R, OCaml, Groovy, F#, Clojure, D, Objective-C, Bash, PowerShell, Vala).
- Add tree-sitter AST backends for C, C++, Scala, Erlang, Elixir, Perl, and Lisp/Scheme/Emacs Lisp, with regex fallback on parse failure.
- Filter Rust restricted visibility: `pub(crate)`, `pub(super)`, `pub(self)`, and `pub(in path)` are no longer treated as exported.
- Add TypeScript CommonJS-style `export = Name` support in both regex and AST backends.
- Update specs/exports/exports.spec.md and tasks.md to match.

## Test Plan
- [x] cargo test — 182 passed, 0 failed
- [x] cargo build --release — clean build
- [x] specsync check — 1 spec passed, 22/22 exports documented

🤖 Generated with [Claude Code](https://claude.com/code)


0xLeif:
## Response to Gemini Code Assist review

Went through each finding individually — fixed the two real ones, and empirically disproved the four node-kind/field-name claims (all four are factually wrong for the tree-sitter grammar versions actually pinned in `Cargo.toml`).

### Fixed
- **`src/validator.rs` (medium)** — switched `fs::read(&full_path)` to `fs::File::open(&full_path)` for the readability check, avoiding loading the whole file into memory just to check it opens.
- **`src/exports/swift.rs` (medium, brace-in-string-literal desync)** — added `mask_string_literals`, a byte-length-preserving pass that blanks string-literal contents (including Swift's `\(...)` interpolation) purely for brace-position lookup in `find_body`/`scan_container_members`, so a `{`/`}` inside a string can no longer desync body-boundary detection. Added a regression test using a dictionary-literal-shaped interpolation payload to prove it.
- **`src/exports/ast/c.rs` (high, hardcoded absolute test path)** — not present in the current code; this must have been from an earlier revision of the branch.

### Verified incorrect (not applied)
For each of the four node-kind/field-name claims below, I wrote a throwaway test that parses a minimal real repro with the actual pinned grammar version and prints every `node.kind()` in the resulting tree (plus, where relevant, whether `child_by_field_name(...)` actually resolves) — rather than relying on recollection of the grammar's shape. All four claims turned out to be wrong for these specific versions:

- **`src/exports/ast/perl.rs` (tree-sitter-perl 1.1.2)** — claimed the node kind should be `subroutine_definition`, not `function_definition`. Parsing `sub foo { return 1; }` produces `(function_definition name: (identifier) body: (block ...))` — `function_definition` is correct, and `extract_exports` returns `["foo"]` as expected, not an empty/fallback result.
- **`src/exports/ast/scala.rs` (tree-sitter-scala 0.26.0)** — claimed it should be `method_definition`/`method_declaration`, not `function_definition`/`function_declaration`. Parsing `class Foo { def bar(x: Int): Int = x + 1 }` produces a `function_definition` node under `template_body`, and extraction correctly returns `["Foo", "bar"]`.
- **`src/exports/ast/erlang.rs` (tree-sitter-erlang 0.19.0)** — claimed `fun_decl` has no field named `"clause"`. Empirically, `child_by_field_name("clause")` resolves correctly even though the child's own *kind* is `function_clause` — tree-sitter allows a child to be both a distinct kind and exposed via a named field simultaneously, which is what's happening here. `-compile(export_all).` extraction works correctly (verified with a two-clause function).
- **`src/exports/ast/lisp.rs` (tree-sitter-commonlisp 0.4.1, "dead code" claim)** — claimed there's no `defun`/`defun_header` node kind, just generic `list_lit`/`sym_lit`. Parsing `(defun foo (x) (+ x 1))` produces a real, dedicated `defun` node with a `defun_header` field child exposing `keyword`/`function_name` — this is the live, correct code path, not dead code.

My best guess is the review ran against different/hypothetical grammar version assumptions rather than the versions actually pinned here — worth keeping in mind for future automated reviews on this repo, since blindly applying any of these four suggested fixes would have broken currently-correct, tested behavior.

Full test suite (1400+ unit tests across all touched files) passes, clippy clean, fmt clean.

0xLeif:
All seven Gemini Code Assist review items are now addressed or already resolved:

1. **Hardcoded absolute paths** — searched the entire `src/exports/` tree; no `/private/tmp/...` or similar absolute paths remain in tests. The `/tmp/` strings that do exist are inline Bash source literals exercising heredoc handling, not Rust test fixtures referencing the local filesystem.

2. **Perl `subroutine_definition` vs `function_definition`** — empirically verified with the pinned `tree-sitter-perl 1.1.2` grammar: `sub foo { ... }` parses as `(function_definition name: (identifier) ...)`. Tests pass.

3. **Scala `method_definition` vs `function_definition`** — empirically verified with `tree-sitter-scala 0.26.0`: `class Foo { def bar = ... }` produces a `function_definition` node under `template_body`. Tests pass.

4. **Erlang `clause` field** — empirically verified with `tree-sitter-erlang 0.19.0`: `child_by_field_name("clause")` resolves correctly even though the child kind is `function_clause`. `-compile(export_all).` extraction works (multi-clause verified).

5. **Swift brace-in-string-literal** — fixed in an earlier commit by `mask_string_literals`; added regression test.

6. **`validator.rs` `fs::read` memory use** — fixed in an earlier commit by switching to `fs::File::open`.

7. **Lisp `defun` dead code** — empirically verified with `tree-sitter-commonlisp 0.4.1`: `(defun foo () ...)` produces a dedicated `defun` node with a `defun_header` field. This is the live path; the fallback `list_lit` arm covers Scheme/Elisp.

Latest run: 1467 unit + 182 integration tests pass; fmt/clippy/spec-check/CI all green.

PR #391, opened 2026-07-17
docs: name the two pre-approval gates in the quickstart

0xLeif:
## What

Adds a callout to Quickstart step 2 naming the **two gates that fire before `change approve` succeeds**:

1. **Artifact completeness** — scaffold `TODO` markers in workspace artifacts fail approval with a path-and-line diagnostic.
2. **Semantic deltas** — every affected spec module needs a `deltas/<module>.md`, and the set must exactly match affected specs.

## Why

The quickstart currently sends readers from `change new` straight to `change approve`. In practice (verified by walking the flow end-to-end on 5.1.0), approval fails twice first — both gates are undocumented at this point in the guide, and both error messages only make sense if you already know what artifacts and deltas are. This is the precise spot where a first-time user stalls.

Wording stays tool-neutral on the delta format itself; #390 documents the format in full, and the two pages cross-reference cleanly (this one doesn't link it to stay mergeable independently).

## Scope

One file, one callout block. No other content changed.

### DIFF
--- a/site/src/content/docs/quickstart.md
+++ b/site/src/content/docs/quickstart.md
@@ -77,6 +77,11 @@ specsync change new "Document and verify the existing authentication module" \
 
 Answer the returned questions, complete its adaptively selected artifacts, and approve the definition before implementation. Agents installed during `init` conduct this interview conversationally.
 
+> **Two gates fire before `change approve` succeeds:**
+>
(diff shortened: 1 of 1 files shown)


0xLeif:
Consolidated into #390. The exact quickstart commit is preserved there alongside the semantic-delta reference, corrected implementation-backed wording, and governed CHG-0049 scope. Closing this duplicate so review and CI continue in one place.

Issue #399, opened 2026-07-19
Tolerate inert 5.0.1 registry.toml stubs


0xLeif:
## Problem

Repos adopted during the 5.0.1 era sometimes carry a placeholder `.specsync/registry.toml` containing only:

```toml
version = 1

[modules]
```

5.0.1 never parsed it. Under 5.1.1, every change operation that resolves a module fails closed with `failed to parse local registry ... while resolving '<module>'` — the 5.1.1 parser requires a `[registry]` table with a non-empty `name`. The stub contains zero information (no module entries), and deleting it is sufficient: module resolution falls back cleanly to `specs/<module>/<module>.spec.md`. But nothing in the tooling detects, warns about, or repairs it.

Hit in two consumer repos during the Trust 1.1.1 rollout (CorvidLabs/rust-server, CorvidLabs/algo-utils-examples), blocking `change reopen` entirely until the stub was removed by hand.

## Proposal

Any of:
- treat the empty `version = 1` stub as absent and fall back to the default module paths,
- emit a targeted diagnostic (`registry.toml is a 5.0.1-era empty stub; delete it or run 'specsync init-registry --force'`) instead of a raw parse error,
- or have the 5.0→5.1 migrate command (if added — see #396) remove/regenerate inert stubs.

Issue #398, opened 2026-07-19
Batch mode for change correct-owner


0xLeif:
## Problem

`change correct-owner` accepts exactly one path per invocation, and each correction requires a fresh approve → verify → accept cycle. When a 5.0.1-era change's acceptance scope covers many production sources whose canonical ownership was never recorded (a common adoption-era gap), re-acceptance under 5.1.1 surfaces the files one at a time: accept fails on file N, correct it, re-approve, re-verify, accept fails on file N+1, repeat.

Observed during the CorvidLabs Trust 1.1.1 rollout:

- CorvidLabs/swift-asa-viewer: **11** sequential corrections
- CorvidLabs/py-algochat: **17**
- CorvidLabs/swift-algotest: **19** (with a guarded script loop; each file verified against the owning module's frontmatter first)
- CorvidLabs/fledge: 2

Every corrected file was already listed in the target module's frontmatter, so the corrections were mechanical — but each iteration re-ran the full verify lane (minutes each).

## Proposal

`change correct-owner <ID> --all-missing` (or a repeated `--path` flag): enumerate the change's acceptance inputs lacking canonical ownership, validate each against the named module's frontmatter, and record all corrections in one audit entry before the single fresh approval + verification + closing approval cycle.

Issue #397, opened 2026-07-19
Archive repair for ledgers with unspecced production source


0xLeif:
## Problem

Under 5.1.1, an archived change whose historical acceptance inputs include production source with no canonical owner fails closed with:

```
error: CHG-0001-...: archived change historical integrity is invalid: legacy accepted change ... requires exactly one distinct valid historical reconstruction, found 0; first reconstruction failure: acceptance input 'scripts/evaluate-agents.ts' is production source without deterministic canonical ownership
```

There is no remediation path: `change reopen` refuses archived changes, and `change correct-owner` requires state `verifying` plus a canonical spec whose frontmatter lists the file — impossible in a repo with zero specs. The repo's contract gate is then permanently red.

Real case: CorvidLabs/agent-findings (spec-less repo, adoption-era archived CHG-0001 covering `scripts/evaluate-agents.ts`). Affected archived ledgers can't be reconstructed because 5.1.1's legacy acceptance-manifest reconstruction demands canonical ownership that never existed in 5.0.1-era records.

## Proposal

One of:
- an `archive-repair` command that re-signs archived evidence under current rules with audited justification, or
- allow `correct-owner` against archived changes (with the same audit trail as reopen corrections), or
- a documented migration that strips/relaxes ownership requirements for archived (immutable) evidence, since archived changes no longer gate delivery.

Issue #396, opened 2026-07-19
Native migrate path for 5.0→5.1 change ledgers


0xLeif:
## Problem

SpecSync 5.1 cannot parse 5.0.1-era change ledgers: reopening records in `approvals.json` lack the required `stale_acceptance_input_digest` / `current_acceptance_input_digest` fields, so `specsync check` fails closed with `missing field 'stale_acceptance_input_digest'` on every historical reopening. The 5.1.x features that supersede the old hand-recorded flow (native `change reopen`, reacceptance without delta replay via `canonical_applied`) make this the mandatory upgrade path, but there is no native migration command — `specsync migrate` only covers v3→v4.

Observed during the CorvidLabs-wide Trust 1.1.1 rollout (22 repos): every repo carrying 5.0.1-era ledgers needed a field-level backfill before 5.1.1 would validate. We shipped a standalone script (`scripts/migrate_specsync_5_1_records.py` in CorvidLabs/trust) that additively backfills the two fields per reopening (stale := embedded `prior_verification.acceptance_input_digest`; current := next acceptance digest in the ledger chain) and self-verifies closing digests — but it belongs upstream.

## Proposal

`specsync migrate 5.0` (or extend `migrate`): idempotent backfill of the 5.1 reopening digest fields with a verification pass over closing digests, safe to run on any 5.0.x ledger. Bonus: have `check` detect the 5.0.1 schema and print the migrate hint instead of a raw serde error.

PR #400, opened 2026-07-19
Fix: trust squash-merged accepted evidence recorded in main history for archival (CHG-0054)

0xLeif:
## Summary

Squash merges discard the original acceptance-transition commits while preserving the accepted evidence bytes in default-branch commits. For an accepted change whose evidence was refreshed while already accepted and then squash-merged, `specsync change archive` failed closed with `requires exactly one trusted transition … found 0` even though a commit on `main` recorded the change as accepted with byte-identical `state.json`/`verification.json`/`approvals.json` — blocking four accepted changes (CHG-0048 musl, CHG-0051, CHG-0052, CHG-0053) from archival. This was hit in production on this repo, which squash-merges all PRs.

## Root cause

`authenticated_accepted_transition` only trusted **first-acceptance transition anchors** (commits whose parent does not record the change as accepted), and the staged-snapshot fallback additionally required the verification commit to be an ancestor of `HEAD`. Squash-merged re-verification produces neither: the re-acceptance commit stays on the discarded branch, and on the default branch the refreshed evidence lands in a commit whose parent is already accepted.

## Fix

When the established searches find no eligible anchor, fall back to **recording anchors**: any commit reachable from `HEAD` or the remote default whose `state.json` records the change as accepted. Every existing per-anchor check is unchanged — committed verification/approvals bytes must equal the current evidence, the committed state must parse as this change in `accepted`, and the record must project exactly onto the committed snapshot. The evidence-key dedupe and the **exactly-one-eligible rule still fail closed** on missing or ambiguous evidence.

Behavior for merge-commit and first-acceptance histories is unchanged: the fallback only runs when the transition-anchor search (and, for archived records, the `accepted-state.json` addition scan) finds nothing.

## Changes

- `src/change.rs` — extract the shared per-anchor evaluation into `consider_accepted_evidence_anchor`, add `accepted_recording_anchors`, and wire the fallback into `authenticated_accepted_transition`.
- Canonical spec `change` v39 → v40: new **REQ-change-037** and Invariant 26 via accepted **CHG-0054** semantic deltas.
- Regression tests: squash-merged evidence refresh while already accepted archives successfully; tampered closing evidence with no matching in-history accepted record still fails closed (negative control verified by disabling the fallback).
- Accepted lifecycle evidence: CHG-0051/52/53 reopened and re-accepted against contract v40; CHG-0048 (release), CHG-0049 (docs/diagnostics), CHG-0050 archived.

## Verification

- `cargo test` — full suite green (1703 lib + 218 integration), run on every lifecycle verify gate (4×)
- `specsync check --strict` — 62/62 specs pass, 100% file and LOC coverage
- `fledge trust verify` — trust gate passed (progressive provenance)

## Follow-ups (post-merge)

- Archive the 5 remaining accepted changes (CHG-0048 musl, CHG-0051, CHG-0052, CHG-0053, CHG-0054) — the archiver refuses while this delivery diff is unmerged (`archive after merge`).
- Related rollout findings filed as #396 (5.0→5.1 ledger migration), #397 (archive repair for unspecced-source ledgers, adjacent to this fix), #398 (batch `correct-owner`), #399 (tolerate inert 5.0.1 registry stubs).


0xLeif (commented):
## Adversarial review (trust model / CHG-0054)

**Verdict: no trust-model blocker — safe to merge from a correctness/trust standpoint**, with should-fix coverage gaps below.

Security review found no medium+ issues. Parent-state was a first-transition selector, not a crypto gate; recording-anchor fallback keeps the same per-anchor authenticity checks (byte equality, accepted identity, projection equality, evidence-key dedupe, exactly-one fail-closed) and closing-digest validation still runs before archive.

### Trust soundness
- Recording path (`src/change.rs` ~8138–8176, 8314–8331) drops only the parent-not-accepted filter; `consider_accepted_evidence_anchor` (~8194–8222) still requires same-commit state/verification/approvals bytes matching workspace evidence + projection equality.
- `found 2` from two commits with identical ver/approvals but distinct raw `state.json` that deserialize equal remains correctly fail-closed (~8363–8368). Availability risk only (serialization drift / `skip_serializing_if` fields); not a forgery bypass.
- Fallback order preserved: transition → archived `--diff-filter=A` → recording (only if `eligible.is_empty()`) → staged. Previously successful transition/archived paths do not downgrade to recording.

### Archived re-validation
- Archive preflight re-runs `validate_archived_integrity` on an archived projection (~3268–3272); post-commit re-validation should land on the archived addition scan with the same evidence bytes. No hole found; missing an explicit post-commit re-validation test.

### Reproduce (spot-checked)
- Targeted squash-refresh + negative tests: green
- Optional negative control (gate recording fallback): positive test fails with `found 0`
- `cargo fmt --check`: clean; no new clippy in the changed region
- `specsync check --strict`: 62/62, 100% coverage
- `fledge trust verify`: trust gate passed
- Archive while unmerged: `archive after merge` (by design)

### Lifecycle evidence
- CHG-0051/52/53/54 approvals + reopenings look chronologically consistent and bound to contract v40; nothing malformed/out of order spotted.

### Should-fix (non-blocking)
1. In `refreshed_accepted_evidence_squash_merged_while_accepted_archives` (~14769), assert transition anchors find no current-evidence match so the test proves the recording path (not staged/transition).
2. Reshape `squash_merged_recording_anchor_fails_closed_without_matching_evidence` (~14851) to the refresh-while-accepted shape; assert `ensure_closing_approval_valid` still `Ok` before archive fails at trusted-transition.
3. Add a recording-path `found 2` case (dual state serialization, identical ver/approvals).
4. Cover recording + nested project root and/or remote-default-only (HEAD without the squash).
5. After archive + commit, re-run `validate_archived_integrity` and confirm the addition-scan path (not only preflight recording).

### Nits
- `#[allow(clippy::too_many_arguments)]` on `consider_accepted_evidence_anchor`
- Pre-existing soft `unwrap_or_else` to `original_state_bytes` at archive (~3260–3262) is awkward but fail-closed via preflight
- CHG-0054 `affected_paths` lists `tests/integration/change.rs` but new coverage is unit-only in `src/change.rs`

PR #449, opened 2026-07-24
fix: hooks/agents managed-block uninstall, real git hooks dir resolution, install backups (closes #415, closes #438)

### LINKED ISSUE #415: hooks uninstall deletes the user's pre-existing pre-commit hook
## Problem

`specsync hooks uninstall` deletes the user's **entire** pre-existing `.git/hooks/pre-commit` file instead of removing only the block `hooks install` appended.

Verified on v5.2.0:

```bash
printf '#!/bin/sh\necho "user hook content"\nexit 0\n' > .git/hooks/pre-commit
chmod +x .git/hooks/pre-commit
specsync hooks install    # appends its block — file now contains user content + specsync block
specsync hooks uninstall  # .git/hooks/pre-commit is GONE — user content lost
```

Markdown companions (AGENTS.md/CLAUDE.md) correctly use managed-block removal; the git hook path just `rm`s the file.

## Suggested fix

Wrap the installed block in begin/end markers (like the markdown companions) and uninstall by removing only the marked block; if the file contains no managed block, leave it untouched. Refuse to delete a hook file the tool didn't create.

Related hooks findings from the same probe (all verified): the installed hook can never block out of the box (default `enforcement=warn` exits 0 on failure); `hooks install` in a non-git dir fabricates `.git/hooks/pre-commit` and reports success; `core.hooksPath` is ignored (installs where git never looks); and the specsync block is appended *after* an existing hook's `exit 0`, making it dead code.

### LINKED ISSUE #438: hooks/agents install lies and destroys — fake hook in nested roots, silent overwrite of user customizations, empty Companion files section
## Problem

`hooks install` and `agents install` lie or destroy in three ways (verified on v5.2.0; sibling of #415, which covers `hooks uninstall` data loss):

1. **Nested project root → fake hook with a green checkmark**: `hooks install` in `repo/packages/app` creates `packages/app/.git/hooks/pre-commit` and prints `✓ Installed Git pre-commit hook`, but git never runs it (real gitdir is `repo/.git`) and git itself ignores the bogus `.git` dir. Protection that doesn't exist, reported as installed. In linked worktrees and submodules (where `.git` is a file) the same code fails loudly with `Failed to create .git/hooks/: Not a directory` — it never resolves `git rev-parse --git-common-dir` in either case.
2. **`agents install` silently overwrites user-customized skill/command files** — no warning, no backup; an appended line is clobbered to stock while output claims `✓ Installed`. (Idempotent healing of deleted/truncated files works and is good; customized files deserve the same managed-block treatment the markdown companions get.)
3. **The installed SKILL.md (byte-identical for all four tools) contains an empty `## Companion files` section** — its content lands under the next heading instead.

Also: `--format json`/`--json` is accepted but ignored on `agents install`/`agents status`; the installed hook can never block out of the box (default `enforcement=warn` exits 0 on failure); install in a non-git dir fabricates `.git/hooks` and reports success; `core.hooksPath` is ignored; and the specsync block is appended after an existing hook's `exit 0`, making it dead code.

0xLeif:
## Problem

**#415 (data loss):** `specsync hooks uninstall --pre-commit` deleted the user's ENTIRE pre-existing `.git/hooks/pre-commit` (a 3-line user hook + appended specsync block = <35 lines starting with `#!/bin/sh` and containing "specsync check" → `fs::remove_file`). Also, the specsync block was appended AFTER an existing hook's `exit 0` — dead code that never ran.

**#438 (lies):** `hooks install` in a nested project root (`repo/packages/app`) created a fake `.git/hooks/pre-commit` (green ✓ but git never runs it); install in a non-git directory fabricated `.git/hooks` and reported success; `core.hooksPath` was ignored; `agents install` silently overwrote user-customized skill files with no warning or backup; the installed SKILL.md had an empty `## Companion files` section.

## Repro (v5.2.0)

```
$ printf '#!/bin/sh\necho "user hook content"\nexit 0\n' > .git/hooks/pre-commit
$ specsync hooks install --precommit   # ✓ block appended AFTER exit 0 (never runs)
$ specsync hooks uninstall --precommit # ✓ Removed
$ cat .git/hooks/pre-commit            # No such file — USER HOOK DELETED
```

After:
```
$ specsync hooks install --precommit   # sentinel block inserted BEFORE exit 0
$ specsync hooks uninstall --precommit # removes only the specsync block
$ cat .git/hooks/pre-commit            # user hook intact (shebang/echo/exit 0)

$ cd packages/app && specsync hooks install --precommit
# → installs into repo/.git/hooks (resolved via `git rev-parse --git-path hooks`)
$ mkdir /tmp/nogit && cd /tmp/nogit && specsync hooks install --precommit
# ✗ /tmp/nogit is not inside a git repository … Run `git init` first … (exit 1, nothing created)
```

## Root cause / fix

- Pre-commit installs are now wrapped in `# >>> specsync:pre-commit … >>>` / `# <<< specsync:pre-commit <<<` sentinels (same managed-block strategy as the markdown companions). Uninstall removes exactly the block — legacy ≤5.2.0 marker blocks and whole-file stock installs are also recognized and removed precisely — and the file is deleted only when nothing but our content remains. A hook with any user content is never deleted; a partial/corrupted marker set errors instead of guessing.
- Install inserts the block BEFORE a trailing `exit 0` instead of appending dead code after it.
- The hooks directory is resolved via `git rev-parse --git-path hooks`, which handles nested project roots, linked worktrees/submodules (`.git` as file), and `core.hooksPath`. Non-git directories now fail loudly with remediation instead of fabricating a `.git/hooks` tree git would never read.
- `agents install` backs up any differing pre-existing file to `<name>.specsync-backup` and prints a warning before overwriting; SKILL.md's `## Companion files` section now actually contains the companion-files guidance (it was an empty heading).

## Scope notes

- The issue's "also" item about `--format json` being ignored on `agents install/status` is a cross-command formatting concern, explicitly out of scope here.
- `agents uninstall` intentionally does NOT delete `.specsync-backup` files (user data).

## Tests

New regression tests — `src/hooks.rs`: `install_precommit_errors_in_non_git_dir`, `install_precommit_wraps_block_in_sentinels`, `install_precommit_inserts_before_trailing_exit_0`, `install_precommit_resolves_nested_project_root_via_git`, `install_precommit_honors_core_hooks_path`, `uninstall_precommit_removes_only_managed_block`, `uninstall_precommit_deletes_file_that_is_purely_ours`, `uninstall_precommit_leaves_user_hook_untouched`, `uninstall_precommit_removes_legacy_appended_block`, `uninstall_precommit_removes_legacy_whole_file_hook`. `src/agents.rs`: `install_backs_up_user_customized_skill`, `install_backs_up_user_customized_command`, `skill_md_companion_files_section_is_not_empty`.

Full suite green (plus the one known pre-existing root-only failure in `change::tests`); `cargo clippy -- -D warnings` clean; all repros re-verified end-to-end with the patched binary.

Closes #415
Closes #438

### DIFF
--- a/src/agents.rs
+++ b/src/agents.rs
@@ -12,6 +12,14 @@ use std::path::{Path, PathBuf};
 
 const SKILL_BODY: &str = r#"## Companion files
 
+Each canonical spec may have policy-selected companion files. Read and update the ones present; do not create empty companions only for ceremony:
+
+- **`tasks.md`** — Work items for this module. Check off tasks (`- [x]`) as you complete them. Add new tasks if you discover work needed.
+- **`requirements.md`** — Acceptance criteria and user stories. These are permanent invariants, not tasks — do not check them off. Update if requirements change.
+- **`context.md`** — Architectural decisions, key files, and current status. Update when you make design decisions or change what's in progress.
+- **`testing.md`** — Test strategy: automated test locations, manual QA checklists, and edge cases/boundary conditions.
+- **`design.md`** *(opt-in)* — Layout, component hierarchy, design tokens, and asset references. Present when `companions.design` is enabled in config.
+
 ## Verified SDD change lifecycle (5.0)
 
 For every meaningful source, test, public documentation, schema, or configuration change:
@@ -29,14 +37,6 @@ For every meaningful source, test, public documentation, schema, or configuratio
 Never invent or self-grant either human approval. If an approved definition changes, its digest
 becomes stale and must be approved again. `specsync check` validates canonical specs plus approved
 active deltas, requirement-to-test evidence, change coverage, and CI gates.
-
-Each canonical spec may have policy-selected companion files. Read and update the ones present; do not create empty companions only for ceremony:
-
-- **`tasks.md`** — Work items for this module. Check off tasks (`- [x]`) as you complete them. Add new tasks if you discover work needed.
-- **`requirements.md`** — Acceptance criteria and user stories. These are permanent invariants, not tasks — do not check them off. Update if requirements change.
-- **`context.md`** — Architectural decisions, key files, and current status. Update when you make design decisions or change what's in progress.
-- **`testing.md`** — Test strategy: automated test locations, manual QA checklists, and edge cases/boundary conditions.
-- **`design.md`** *(opt-in)* — Layout, component hierarchy, design tokens, and asset references. Present when `companions.design` is enabled in config.
 
 ## Before modifying any module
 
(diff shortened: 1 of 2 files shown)


0xGaspar (approved):
Approving. This is the strongest PR in the current batch.

**Verified:**
- Merges **cleanly** into current `main` (no conflicts).
- I merged it together with #455, #456 and #457 and ran the suite on the result: **2070 unit + 335 integration tests, 0 failures.**
- Scope is tight and matches the description exactly: only `src/hooks.rs` and `src/agents.rs` change vs the branch point.

**What I like:** the sentinel managed-block approach is the right call over line-count heuristics, and "a hook with any user content is never deleted; a partial/corrupted marker set errors instead of guessing" is exactly the correct failure posture for a data-loss bug. Resolving the hooks dir via `git rev-parse --git-path hooks` rather than assuming `.git/hooks` correctly handles worktrees, submodules and `core.hooksPath` in one move. Failing loudly in a non-git directory instead of fabricating a tree git will never read is a genuine honesty fix, not just a bug fix.

**Non-blocking notes:**

1. `fmt` is red on this PR in CI. Worth a `cargo fmt` pass before merge — the code change itself looks fine, so I assume this is incidental.
2. `trust`, `spec-check` and `Required CI gate` are red here, but they are red on **every** open PR in this repo right now including the ones that merge and test clean. That looks like a repo-level gate problem rather than anything this PR did — flagging so it isn't mistaken for a defect here.
3. Not deleting `.specsync-backup` files on `agents uninstall` is the right call. Consider mentioning it in user-facing output so people know the backups are there.

Good to merge once `fmt` is green.

---

0xLeif:

PR #446, opened 2026-07-23
Fix: harden MCP security and GitHub issue boundaries

0xLeif:
## Summary

- make MCP read-only by default; mutating tools require `mcp --allow-write` and always use the configured server root
- confine reads beneath retained root capabilities and reject traversal, symlink/junction/reparse, special-file, identity-replacement, malformed-input, and notification-write paths
- fail closed across checked Gradle/Cargo/Node discovery, CLI coverage, GitHub issue/import paths, response/input budgets, and exact JSON-RPC argument validation
- synchronize canonical specs, semantic deltas, companions, lifecycle artifacts, changelog, and public security documentation

## Exact revision

- final implementation and accepted evidence: `044ab6f46518d6ed8d61287a22ca279eda0ad069`
- final source implementation parent: `dec4cd2f1849b05fafbdcecfcb428db59b4ce022`

## Verification

- formatting, Clippy/type checks, release build, and complete `fledge lanes run verify`: passed
- unit tests: **1,959 passed, 0 failed**
- integration tests: **319 passed, 0 failed**
- lifecycle: **68 active changes valid**
- spec validation: **62/62 passed**, zero warnings or failures
- coverage: **105/105 files** and **107,701/107,701 LOC** (**100%**)
- spec quality: **62/62 at 100/A**
- CHG-0063 and dependent CHG-0062 were freshly verified in final dependency order and accepted using the human closing approval
- no unresolved inline review threads; the independent review's only medium finding is fixed
- native hosted Windows junction/reparse tests passed on the final source implementation tree
- exact-tree private `CorvidLabs/spec-sync-sandbox` replay passed with hash-bound evidence
- `fledge trust verify --range origin/main..HEAD`: passed
- Augur: **review**, risk **51.18**, below the block threshold; no BLOCK verdict
- Attest: commit note recorded and verified with tests passed and human approval

## Hosted gate

The final GitHub Actions run for `044ab6f` is in progress. Merge is gated on all required hosted checks succeeding.

Fixes #414

PR #450, opened 2026-07-24
fix: confine MCP tool roots to the server root (closes #414)

### LINKED ISSUE #414: MCP tools accept arbitrary root paths — arbitrary file write outside the server root
## Problem

Every MCP tool accepts a `root` argument that is used verbatim — there is no confinement to the server's `--root`. A caller (or a prompt-injected agent client) can make the server write files anywhere on the filesystem.

Verified on the published v5.2.0 asset:

```
$ specsync mcp --root /tmp/legit-server
# JSON-RPC: tools/call specsync_generate {"root": "/tmp/victim"}
# → {"count": 1, "generated": ["specs/evil/evil.spec.md"]}
# /tmp/victim/specs/evil/evil.spec.md now exists, written by the server
```

Any writable path works; nothing checks that `root` equals (or is contained in) the server's configured root.

## Why it matters

The MCP server is explicitly offered to AI agents (`specsync mcp`). An unconfined write primitive turns a prompt-injection into arbitrary file creation/overwrite on the host — well beyond the read-mostly surface (`check`/`coverage`/`score`) the other tools expose.

## Suggested fix

Default every tool to the server's `--root`; if a `root` arg is kept, reject paths that are not the server root or a child of it (canonicalized, symlink-resolved). Write-capable tools (`generate`, anything else that mutates) should be opt-in via a server flag.

Related hardening from the same probe: tool args aren't validated against their advertised schemas (`strict: "yes please"` coerces to false without `-32602`), and known-method notifications (no `id`) incorrectly get a `"id":null` response, which JSON-RPC forbids.

Reproduced via stdio JSON-RPC probe; happy to share the script.

0xLeif:
## Summary

Every MCP tool accepted a caller-supplied `root` verbatim, allowing an arbitrary file write outside the server root. **Verified exploit:** `specsync_generate` with `{"root":"/tmp/victim"}` wrote `/tmp/victim/specs/evil/evil.spec.md` — a hostile or prompt-injected MCP client could make the server process write files anywhere it had permissions.

This was handled per SECURITY.md. Given the arbitrary-write primitive, we recommend maintainers consider publishing a security advisory for this issue.

## Fix

- **All tools default to the server `--root`.** A caller-supplied `root` is accepted only if it is the server root or a path inside it; anything else is rejected with `-32602` (Invalid params) via the new `util::confine_path_to_root`, which canonicalizes both paths (symlink- and `..`-resolved) before comparing.
- **Write tools gated:** `specsync_generate` and `specsync_init` now require the new `specsync mcp --allow-writes` flag, so a connected agent gets a read-only surface by default.
- **Schema validation:** `tools/call` arguments are validated against the schemas advertised by `tools/list`; unknown argument names and wrongly typed values are rejected with `-32602` before touching the filesystem.
- **JSON-RPC correctness:** notifications for known methods (no `id` member) no longer receive a forbidden `"id": null` response; parse errors and explicit `"id": null` requests still behave per spec.

## After

The same call — `specsync_generate {"root":"/tmp/victim"}` — now returns `-32602 path ... is outside the server root` and writes nothing.

## Tests

- New `util::confine_path_to_root` unit tests (child accepted, sibling/`..`/symlink escape rejected, nonexistent rejected) and `mcp.rs` unit tests covering rejection, write gating, arg-type validation, and notification handling.
- New integration tests: `mcp_tools_call_rejects_root_outside_server_root`, `mcp_write_tools_require_allow_writes_flag`, `mcp_tools_call_rejects_wrongly_typed_arguments`, `mcp_known_method_notification_gets_no_response`.
- Full suite passes: 1729 unit + 227 integration tests, clippy clean.

Closes #414

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -107,7 +107,12 @@ pub enum Command {
     /// Watch spec and source files, re-running check on changes
     Watch,
     /// Run as an MCP (Model Context Protocol) server over stdio
-    Mcp,
+    Mcp {
+        /// Enable write-capable MCP tools (specsync_generate, specsync_init).
+        /// Off by default so a connected agent gets a read-only surface.
+        #[arg(long)]
+        allow_writes: bool,
+    },
     /// Scaffold a new spec with required companion files and optional design.md
     AddSpec {
         /// Module name for the new spec
--- a/src/main.rs
+++ b/src/main.rs
@@ -160,7 +160,7 @@ fn run() {
             &cli.only_status,
         ),
         Command::Watch => watch::run_watch(&root, cli.strict, cli.require_coverage),
-        Command::Mcp => mcp::run_mcp_server(&root),
+        Command::Mcp { allow_writes } => mcp::run_mcp_server(&root, allow_writes),
         Command::AddSpec { name } => commands::scaffold::cmd_add_spec(&root, &name),
         Command::Scaffold {
             name,
--- a/src/mcp.rs
+++ b/src/mcp.rs
@@ -13,7 +13,12 @@ const SERVER_VERSION: &str = env!("CARGO_PKG_VERSION");
 const PROTOCOL_VERSION: &str = "2024-11-05";
 
 /// Run the MCP server on stdio.
-pub fn run_mcp_server(root: &Path) {
+///
+/// All tool `root` arguments are confined to the server `root` (see
+/// [`crate::util::confine_path_to_root`]); write-capable tools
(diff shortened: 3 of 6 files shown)


0xLeif:
Closing this PR as superseded by #446. An independent exact-diff comparison found no unique production code, test, specification, or acceptance behavior here that is not already covered by #446. Keeping #446 as the single delivery path avoids merging duplicate MCP security implementations.

PR #451, opened 2026-07-24
fix: cache check outcomes so warm-cache runs keep reporting warnings (closes #429)

### LINKED ISSUE #429: Non-strict check is run-history dependent — warnings vanish on cache-warm reruns; JSON reports specs_checked:0
## Problem

Non-strict `specsync check` is run-history dependent: on an unchanged repo the first run prints full per-spec results including warnings; every later run prints only `⊘ Skipped N unchanged spec(s) / All specs unchanged — nothing to validate` — **warnings are reported once and then vanish silently**.

The JSON variant is worse: a warm-cache `check --format json` returns

```json
{ "warnings": [], "specs_checked": 0, "passed": true }
```

so a CI JSON consumer sees zero specs checked and zero warnings on every rerun after the first. Verified on v5.2.0 with output diffed byte-for-byte across 20 runs.

Related cache findings from the same rounds: the hash cache itself is effectively inert (forged-hash poisoning changes nothing; `--force`/`--no-cache` behave identically in output and timing even at 3000 exports), and `rehash` doesn't write the `.specsync/config.toml` cache entry `check` requires, so the first `check` after `rehash` re-validates everything anyway.

## Suggested fix

Cache outcomes, not just file hashes: a warm-cache run should report the same warnings (marked "cached"/"unchanged") in text and JSON, with `specs_checked` reflecting the full set or an explicit `specs_skipped` count. Include the config entry in `rehash`'s cache writes.

0xLeif:
## Problem

Non-strict `check` was run-history dependent. After a first run reported warnings and populated the hash cache, a warm-cache rerun printed "All specs unchanged — nothing to validate" and `--format json` returned `{warnings: [], specs_checked: 0, passed: true}` — warnings were reported once and then silently vanished, even though nothing was fixed. Additionally, `rehash` did not write the config cache entry, so the first `check` after a `rehash` always saw a "changed" config and re-validated everything anyway.

## Fix

- **`HashCache.outcomes`**: new map of per-spec `CachedSpecOutcome` (errors/warnings/notices), written alongside the hashes. Backwards-compatible — old caches without the field still parse (`serde(default)`).
- **`run_validation` records outcomes** for every spec it validates.
- **`check` replays cached findings** for unchanged (skipped) specs, marked with a `(cached)` suffix in both text and JSON output, and counts them in totals, `passed`, and the exit gates — so a warm run reports the same warnings as the cold run that produced them. JSON output also gains `specs_skipped`.
- **`rehash` writes the global-input entries** (config file + schema files) that `check` consults, making the post-rehash run genuinely warm.

## After

Warm `check --format json` now returns `specs_checked` reflecting the full spec set, `specs_skipped` for the cache-served specs, and the same warnings as the cold run, each marked `(cached)`.

## Tests

- `warm_cache_json_replays_cached_warnings`, `warm_cache_text_replays_cached_warnings`, `rehash_writes_config_cache_entry` (integration)
- `outcomes_round_trip_and_default_for_old_caches` (unit, `hash_cache`)
- Full suite passes: 1729 unit + 227 integration tests, clippy clean.

Closes #429

### DIFF
--- a/specs/cmd_check/cmd_check.spec.md
+++ b/specs/cmd_check/cmd_check.spec.md
@@ -1,11 +1,11 @@
 ---
 module: cmd_check
-version: 8
+version: 9
 status: stable
 files:
   - src/commands/check.rs
 db_tables: []
-tracks: []
+tracks: [429]
 depends_on:
   - specs/commands/commands.spec.md
   - specs/comment/comment.spec.md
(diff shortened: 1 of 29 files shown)


0xGaspar (changes requested):
Requesting changes — rebase only; I have no objection to the code.

**Conflicts with `main`** on `src/commands/check.rs`, `src/commands/mod.rs` and eight spec files, from the shared stale base described below.

**On the substance this is an important fix and I'd like it to land.** "Warnings reported once, then silently vanish on the next run, even though nothing was fixed" is a genuinely dangerous property for a validation tool — it means a warm CI cache can turn a failing check green without anyone touching the code. Making `check` run-history-independent is the right invariant, and it's worth stating in the spec in exactly those terms: *the result of `check` must not depend on whether the cache is warm.*

Specifics I like:

- `serde(default)` on the new `outcomes` map so old caches still parse — correct, and avoids forcing a cache wipe on upgrade.
- Marking replayed findings `(cached)` in both text and JSON rather than making them indistinguishable. Users should be able to tell a replayed finding from a freshly computed one.
- Counting cached findings in totals, `passed` **and the exit gates**. That last one is the part that actually fixes the bug; the display change alone wouldn't have.
- `rehash` writing the global-input entries so the next `check` is genuinely warm. Nice catch that the previous behaviour made `rehash` almost pointless.

Two questions, neither blocking:

1. **Cache invalidation on tool upgrade.** If a future release changes what a given check reports, cached outcomes from an older binary would replay stale findings. Is there a version stamp in the cache keyed to the specsync version, or should there be?
2. **`specs_skipped` in JSON** is a good addition — is it in the documented schema for `check --format json`, and does adding it need a `schema_version` bump under this repo's envelope conventions?

Rebase and I'll approve.

---

---

0xLeif:

Issue #442, opened 2026-07-21
new --full produces a scaffold that fails the tool's own validation (7 scaffold-marker errors, greedy auto-detect)


0xLeif:
## Problem

`specsync new --full` produces a scaffold that fails the tool's own validation — strictly worse than plain `new` (verified on v5.2.0):

1. **7 "Unfilled companion scaffold marker" warnings** on the immediate `check --strict` (promoted to errors in strict mode) — while plain `new` passes with 0 warnings.
2. **Still omits the required `## Invariants` / `## Behavioral Examples` / `## Error Cases` sections**; `status: draft` merely hides it — promoting to `active` fails check, and `score` deducts for them.
3. **Greedy auto-detect with no name matching**: `new ghost --full` claimed the only existing file `src/gadget.rs`, producing "duplicate spec ownership" check errors across both specs, with no warning at creation time.
4. With zero source files it warns and writes `files: []`, but `check` then reports a vague `✗ Frontmatter invalid` (the real reason is buried in Suggested fixes).

Related name-validation gaps from the probe round: 200-char names accepted (300 chars dies with raw `os error 63` at `mkdir`); reserved names (`change`, `specs`, `con`) accepted; `Lib` vs `lib` is rejected only on case-insensitive filesystems with a wrong-path error message (Linux CI would mint the collision); names with spaces/leading dashes/emoji accepted; and `new --help` documents no naming rules at all.

PR #452, opened 2026-07-24
fix: import --from-dir preserves source content instead of discarding it (closes #416)

### LINKED ISSUE #416: import --from-dir discards source content and writes skeleton templates that fail their own gate
## Problem

`specsync import --from-dir` does not import content — it replaces every markdown file with a fresh skeleton template, and every file gets a ✓ as if it imported cleanly.

Verified on v5.2.0: a hand-written spec with frontmatter (`module/version/status/files`) and real body text goes in; the output spec contains none of it (zero occurrences of the original purpose text), and its `files:` section is empty.

Two follow-on failures:

1. Imported specs write `files: []`, which then fails the tool's own gate on the very next `check --strict` with `Frontmatter missing required field: files (must be a non-empty list)` — 100% of imports fail immediately.
2. Change Log attribution is wrong: `--from-dir` imports are stamped `Imported from Confluence:` even though Confluence was never involved.

Also broken: the error for GitHub repo auto-detect tells users to set `github.repo` in `specsync.json` — a file the tool never reads (the working key is `[github] repo` in `.specsync/config.toml`).

## Expected

Parse the source file: preserve frontmatter fields, sections, API tables, and changelogs; only scaffold what's missing. A file that can't be parsed should fail loudly, not be silently overwritten with a ✓.

0xLeif:
## Problem

`specsync import --from-dir docs/` replaced every markdown file with a Confluence-shaped skeleton: frontmatter, body sections, API tables, and changelogs were silently discarded while the command printed `✓` for each file. Imported specs were written with `files: []` (failing the tool's own non-empty-files gate on the next `check`), the Change Log was stamped `Imported from Confluence:` regardless of source, and the GitHub error message pointed at `specsync.json`, a file the tool never reads (the working key is `[github] repo` in `.specsync/config.toml`).

**Before** (v5.2.0): a real spec `docs/auth.md` (module auth, version 2, Purpose + Change Log) imported via `--from-dir` came out as a generic skeleton: `version: 1`, Purpose = `Add user authentication`-style boilerplate, Change Log = `Imported from Confluence: docs/auth.md`, original sections gone — and `✓` printed.

**After**: the same file imports with `version: 2`, original Purpose text and the `2024-01-01 | Real history` changelog row preserved verbatim, missing required sections (Invariants, Error Cases, …) scaffolded underneath, and no `Confluence` stamp.

## Root cause

`cmd_import_from_dir` ran every file through `parse_markdown_as_import_item`, which merely *guessed* a purpose line and fed the Confluence-shaped `importer::render_spec` skeleton, throwing the source content away and hard-stamping `Imported from {source_type}`.

## Fix

- Parse the source file instead: files with spec frontmatter are kept verbatim (including unknown/custom frontmatter fields), patched only where `module`/`version`/`status` are missing or `files:` is empty; plain markdown is wrapped in frontmatter with its body preserved below.
- Source files are auto-detected (`find_files_for_module` + single-source fallback, root-relative) so imported specs pass the non-empty `files:` gate; when nothing is detected a loud warning is printed instead of a quiet `✓`.
- Only missing required sections are scaffolded; the Change Log is stamped `Imported from <path>`.
- Unparseable input (empty file, broken frontmatter) fails loudly per file and exits nonzero — no more silent `✓`.
- GitHub repo error message now names the real config location.

## Tests

`from_dir_preserves_existing_spec_content`, `from_dir_wraps_plain_markdown_without_losing_content`, `from_dir_detects_source_files`, `from_dir_fails_loudly_on_unparseable_input`, `from_dir_strips_spec_suffix_from_stem` — full suite green (1736 passed; the single failure is the pre-existing root-only `change::tests::non_git_walk_skips_volatile_trees_but_fails_on_relevant_walk_errors`), `cargo clippy -- -D warnings` green.

Closes #416

### DIFF
--- a/specs/cmd_import/cmd_import.spec.md
+++ b/specs/cmd_import/cmd_import.spec.md
@@ -1,6 +1,6 @@
 ---
 module: cmd_import
-version: 3
+version: 4
 status: stable
 files:
   - src/commands/import.rs
@@ -17,7 +17,7 @@ depends_on:
 
 ## Purpose
 
-Implements the `specsync import` command. Imports specs from external systems (GitHub Issues, Jira, Confluence) by fetching remote data and converting it into spec files with companions.
+Implements the `specsync import` command. Imports specs from external systems (GitHub Issues, Jira, Confluence) and local Markdown directories. Directory imports preserve complete valid specs byte-for-byte, augment incomplete documents without discarding their content, and reject malformed frontmatter before writing output.
 
 ## Public API
 
@@ -34,6 +34,9 @@ Implements the `specsync import` command. Imports specs from external systems (G
 3. Creates spec and companion files (tasks.md, context.md, requirements.md, testing.md); design.md is generated only when `companions.design` is enabled in config
 4. Will not overwrite existing spec
 5. Success guidance tells users to validate and complete imported details, not to fill template markers
+6. Directory imports never report success for output with an empty `files` list: source code is auto-detected, with the canonically confined project-relative input document used as the ownership fallback
+7. A complete valid spec keeps its original bytes and declared `module`; incomplete content is only supplemented with missing frontmatter fields and required sections
+8. Duplicate or wrongly shaped known frontmatter fields fail the affected import without creating its spec
 
 ## Behavioral Examples
 
@@ -43,6 +46,12 @@ Implements the `specsync import` command. Imports specs from external systems (G
 - **When** `cmd_import` runs
 - **Then** fetches issue #42, creates spec from its title and body
 
+### Scenario: Import an existing spec
+
+- **Given** `docs/renamed.spec.md` is a complete valid spec declaring `module: auth`
+- **When** `specsync import --from-dir docs` runs
+- **Then** creates `specs/auth/auth.spec.md` with bytes identical to the source document
+
 ## Error Cases
 
 | Condition | Behavior |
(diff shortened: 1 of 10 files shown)


0xGaspar (changes requested):
Requesting changes — rebase only; the code looks right to me.

**Conflicts with `main`** on `src/commands/import.rs` and six spec files, from the shared stale base described below.

**On the substance:** this is a data-loss bug and the description doesn't oversell it. `import --from-dir` replacing every markdown file with a Confluence-shaped skeleton while printing `✓` per file is about the worst combination available — destructive and reassuring at the same time. Parsing the source instead of guessing at it is obviously correct.

Things I specifically like:

- Preserving **unknown/custom frontmatter fields** verbatim. Easy to overlook, and dropping fields you don't recognise is its own quiet data loss.
- Failing loudly per file on unparseable input and exiting non-zero, rather than `✓`. Same theme as the rest of this batch: the old behaviour's real defect was the success signal, not just the wrong output.
- Dropping the hardcoded `Imported from Confluence:` stamp for the actual source path. Stamping every import with the wrong provenance is a small lie that compounds.
- Fixing the error message that pointed at `specsync.json`, a file the tool never reads. Pointing users at a nonexistent config file is a genuinely costly bug for the time it wastes.

Two questions:

1. **Is this idempotent?** Running `--from-dir` twice over the same directory — does the second run re-detect `files:`, re-scaffold missing sections, and append a second `Imported from <path>` changelog row? A test for double-import would be reassuring given this command now writes into files it previously overwrote wholesale.
2. **Source auto-detection via `find_files_for_module` + single-source fallback** — how loud is the single-source fallback? #442 flags the same greedy-fallback pattern elsewhere as a bug worth warning about. Worth making sure the two PRs agree on the behaviour.

Rebase and I'll approve.

---

---

0xLeif:

Issue #437, opened 2026-07-21
Export extraction misses across backends — JS \u escapes, Unicode NFC/NFD, indented tables, Erlang/Lisp/Scala/C-header/Perl


0xLeif:
## Problem

Export extraction misses or mis-reports real exports across most non-Rust/TS backends (verified on v5.2.0):

- **JS/TS**: legal `\u` escape identifiers (`export const caf\u0061 = 4`) are truncated at the backslash — extractor reports a bogus export `caf`, and the true name is an unfixable phantom.
- **Unicode**: no normalization in export matching — NFC source `naïve` vs NFD spec `naïve` yields phantom + undocumented errors on glyph-identical names.
- **Public API tables**: an industry-standard 2-space-indented markdown table (valid GFM, renders on GitHub) is silently ignored → `0/N exports documented` with no "table not found" hint; duplicate table rows are never flagged.
- **Erlang**: `-export([area/2])` arity is stripped invisibly — the export is missing while documenting `area/1` is itself phantom-flagged, so no workaround exists.
- **Common Lisp**: the `defpackage` export list is ignored (non-exported internals demanded as public API) while `defgeneric`/`defclass` are never extracted.
- **Scala 3**: `inline def`/`transparent inline def` invisible (and missing from the coverage denominator); a `compare` override inside an anonymous `given Ordering[Point] with` is falsely reported as a top-level export.
- **C headers**: prototypes only — struct/enum typedefs, extern globals, and function-pointer typedefs are uncounted, inflating "N/N documented".
- **Perl**: adding `our @EXPORT_OK = qw(helper_fn);` flips the backend into Exporter-list mode and phantom-errors real OO methods (`new`, `parse` → `0/1 documented`); `new parser` silently misses `src/Parser.pm` (case-sensitive module matching).

Only the Elixir backend came out fully clean.

Issue #435, opened 2026-07-21
db_tables silently unvalidated by default; schema_pattern vacuous; rename/drop inverted; quoting false-positives


0xLeif:
## Problem

`db_tables` validation is silently disabled by default and broken when enabled (verified on v5.2.0):

1. **Silent no-op by default**: `specsync init` doesn't emit `schema_dir`, so declared tables are never validated — phantom tables pass `check --strict` with exit 0 and no hint that validation was skipped.
2. **`schema_pattern` kills validation entirely**: any value (`*.sql`, `**/*`, regex) makes every declared table pass vacuously, including `zzz_definitely_missing`.
3. **Rename/DROP handling inverted**: `ALTER TABLE x RENAME TO y` makes the tool accept the old name and reject the new one; `DROP TABLE`'d tables are reported as existing (the regexes exist in the binary but aren't applied to the existence check).
4. **Standard SQL quoting false-positives**: `"quoted"`, `` `backtick` ``, and `schema.table` names are all reported missing (extraction is `(\w+)`-only); table-name comparison is case-sensitive.
5. **Cache + exit footguns**: editing `schema_dir` in config does not invalidate the hash cache (stale greens without `--force`); a nonexistent `schema_dir` passes silently; default `check` prints "1 failed" but exits 0 (JSON: `"passed": true` with non-empty `errors`) — CI must use `--strict`.

What works: flat `*.sql` plain/`IF NOT EXISTS`/multiline `CREATE TABLE` detection with a configured `schema_dir`.

0xLeif:
Revalidated on current `origin/main` (`fbd87f9e`). Focused `cargo test schema -- --test-threads=1` passed 41 unit + 5 integration tests. Current coverage includes: explicit warning when `db_tables` is declared without `schema_dir`; invalid/non-capturing `schema_pattern` diagnostics; fail-closed nonexistent/unreadable schema paths; rename/drop replay; quoted, qualified, and case-normalized names; schema-input cache invalidation; and strict exit behavior. Closing as fixed.

PR #453, opened 2026-07-24
fix: frontmatter parsing leniency enables validation bypass (closes #436)

### LINKED ISSUE #436: Frontmatter parser leniency is a validation bypass — duplicate keys last-wins; depends_on entries never validated; no path confinement
## Problem

Frontmatter parsing is lenient to the point of validation bypass (verified on v5.2.0):

1. **Duplicate keys silently accepted, last-wins** — a hidden second `status: draft` after `status: active` disables all section/export validation while the file still reads "active" to a human. This is a validation bypass by accident or by malice.
2. **`depends_on` entries are never validated** — `depends_on: [nosuchmod]` passes `check --strict` with `✓ All dependency specs exist`, and `deps` reports `Edges: 0` plus `✓ All dependency declarations are valid` even with real deps declared (the flow-style `[b]` form `specsync new` scaffolds is itself parsed as zero edges).
3. **No path confinement** — `/etc/passwd` and `../` traversal entries validate as ✓ in `resolve` and `check --strict`; `deps` silently ignores the absolute path. Three commands give three different verdicts on the same list.
4. **`module:` is never cross-checked** against the spec's directory/filename (`module: totallydifferent` in `specs/lib/lib.spec.md` passes).
5. **No YAML type/shape validation** — `version: "one"`, `depends_on: {a: b}`, unclosed `[unclosed` / `"mymod`, tabs, and colon-less garbage lines all report `✓ Frontmatter valid`; the only hard failure (missing closing `---`) gives no diagnostic.

## Suggested fix

Reject duplicate frontmatter keys (or at minimum warn loudly and honor the first); validate `depends_on` targets and shapes at check time with one verdict across check/deps/resolve; confine spec paths to the project; cross-check `module:` against the spec path; and surface a warning for non-string versions and non-list dependency fields.

0xLeif:
## Problem

Several frontmatter-parsing leniencies let malformed or hostile specs silently bypass validation:

- **Duplicate keys**: a hidden second `status: draft` silently disabled all section/export checks (last value won, no diagnostic).
- **Flow-style lists**: `depends_on: [b]` (the exact form `specsync new` scaffolds) did not parse, so dependency edges silently vanished.
- **Wrong shapes**: `depends_on: {a: b}` (mapping) parsed to an empty list; unclosed `[` and unterminated quotes were dropped silently.
- **Non-numeric `version` and garbage lines** were accepted without any signal.
- **Dependency verdicts diverged**: `check`, `deps`, and `resolve` each judged `depends_on` entries differently — absolute paths and `..` escapes were probed against the host filesystem by some commands and ignored by others; unreadable specs/sources were silently dropped from the dependency graph.
- **`module:` vs filename**: a spec declaring a module name nothing else can reference produced no diagnostic.

## Repro (fixture: `/var/tmp/scratch/w2/436/`)

```yaml
---
module: dup
module: other        # duplicate key
version: "abc"       # non-numeric
status: active
depends_on: [../outside, legit-mod]   # flow-style + escape
files: src/a.ts
---
```

- **Before**: parses cleanly; zero edges; `../outside` never reported; duplicate `module:` invisible.
- **After**: hard errors for the duplicate key and the `../outside` escape (path confinement), "Dependency spec not found: legit-mod" verdict shared by `check`/`deps`/`resolve`, and warnings for the non-numeric version.

## Root cause

`parse_frontmatter` was a lossy best-effort parser with no error channel, and each command re-implemented (differently) how a `depends_on` entry should be judged.

## Fix

- `parser.rs`: `ParsedSpec` gains `errors`/`warnings`; duplicate-key rejection with offending line; flow-style list parsing (hard error on unclosed bracket/unterminated quote); mapping/scalar shape handling for list fields; version-shape and garbage-line warnings; leading-BOM tolerance.
- `validator.rs`: surfaces parser errors/warnings in the validation result; single `validate_local_dependency` verdict (cross-project refs skipped, escapes rejected, bare module names resolve via specs_dir, path confinement via `source_within_root`); module↔filename cross-check warning.
- `deps.rs` / `commands/resolve.rs`: use the same `validate_local_dependency` verdict; unreadable specs and sources are hard errors instead of silently dropped graph nodes/imports.

`test_parse_frontmatter_duplicate_key_rejected_loudly`, `test_parse_frontmatter_flow_style_string_lists`, `test_parse_frontmatter_rejects_bad_list_shapes`, `test_parse_frontmatter_warns_on_non_numeric_version_and_garbage_lines`, `test_validate_spec_duplicate_status_key_is_an_error`, `test_validate_spec_rejects_dependency_escape_and_flow_style_missing`, `test_validate_local_dependency_rejects_escapes`, `test_validate_local_dependency_one_verdict`, `test_validate_spec_bare_dep_name_resolves_via_specs_dir`, `test_validate_spec_module_filename_mismatch_warns`, `test_flow_style_depends_on_counts_edges`, `test_deps_rejects_escapes_and_missing_with_same_verdict_as_check`.

Full suite: 1743 passed, 1 pre-existing root-only failure (`change::tests::non_git_walk_skips_volatile_trees_but_fails_on_relevant_walk_errors`, unrelated). `clippy --all-targets -- -D warnings` clean.

## File overlap note

`src/parser.rs` and `src/validator.rs` also contain the #435 (schema validation) and #437 (export extraction) changes; those PRs (#fix/435-db-schema-validation, #fix/437-export-extraction) ship their own files and note this overlap. The whole current file is included here since #436 most owns both files.

Closes #436

### DIFF
--- a/specs/cmd_resolve/cmd_resolve.spec.md
+++ b/specs/cmd_resolve/cmd_resolve.spec.md
@@ -1,6 +1,6 @@
 ---
 module: cmd_resolve
-version: 3
+version: 4
 status: stable
 files:
   - src/commands/resolve.rs
@@ -30,7 +30,7 @@ Implements the `specsync resolve` command. Resolves dependency references — lo
 ## Invariants
 
 1. Classifies deps as local vs cross-project
-2. Local verified by file existence
+2. Local references use `validate_local_dependency`: bare modules resolve to canonical spec files and absolute/traversal/backslash/symlink escapes are rejected without host-filesystem probing
 3. No network calls without `--remote` or `--verify`
 4. `--verify` implies `--remote` — registry check always runs first
 5. Warnings for unresolvable refs (does not exit non-zero)
@@ -88,7 +88,7 @@ Implements the `specsync resolve` command. Resolves dependency references — lo
 
 | Condition | Behavior |
 |-----------|----------|
-| Local dep missing | Warning printed |
+| Local dep missing or unconfined | Shared dependency failure printed with the offending entry |
 | Remote registry fetch fails | Warning, continues |
 | Remote spec fetch fails | Warning, continues |
 | Remote spec unparseable | Warning, continues |
@@ -118,6 +118,7 @@ Implements the `specsync resolve` command. Resolves dependency references — lo
 
 | Date | Change |
 |------|--------|
+| 2026-07-26 | v4: use the same confined local-dependency verdict as check and deps, including malformed remote lookalikes |
 | 2026-04-10 | v2: Added `--verify` for deep content verification, `--cache-ttl`, drift detection, exit codes |
 | 2026-04-09 | Initial spec |
 | 2026-07-11 | CHG-0010-canonicalize-every-specsync-5-0-contract-and-requirement: Canonicalize every SpecSync 5.0 contract and requirement |
--- a/specs/cmd_resolve/context.md
+++ b/specs/cmd_resolve/context.md
@@ -9,6 +9,7 @@ spec: cmd_resolve.spec.md
 - **Drift vs warning**: deprecated/removed/archived status and missing consumed exports are breaking `DRIFT` (exit 1); non-bidirectional deps, fetch failures, and parse failures are non-fatal `WARN`.
 - **Cache to avoid re-fetching**: remote spec bodies are cached on disk under `.specsync-cache/remote-specs/` with a TTL (default 3600s); repo/path slashes are sanitized into the cache filename. TTL 0 disables caching.
 - **Consumed-export discovery is table-driven**: `find_consumed_exports` scans the local spec's `### Consumes` table for rows whose module column matches the remote module, extracting backtick-wrapped identifiers.
+- **Shared local verdict**: local and malformed-remote-lookalike entries call `validate_local_dependency`; resolve never performs an unconstrained `root.join(dep).exists()` probe.
 
 ## Files to Read First
 
--- a/specs/cmd_resolve/requirements.md
+++ b/specs/cmd_resolve/requirements.md
@@ -12,8 +12,9 @@ spec: cmd_resolve.spec.md
 
 ## Acceptance Criteria
 
-- `cmd_resolve(root, remote, verify, cache_ttl)` scans every spec's `depends_on`, classifying each ref as local (path) or cross-project (`owner/repo@module`).
-- With no flags, local refs are checked by file existence and cross-project refs are listed only — **no network calls** are made.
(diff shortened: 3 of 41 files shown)


0xGaspar (changes requested):
Requesting changes. There's a specific issue here beyond the rebase, and I'd like it looked at.

**Blocking: `#[allow(dead_code)]` is masking a dropped call site.**

This branch's `src/schema.rs` is near-identical to #455's — they differ by exactly **two lines**, and both are `#[allow(dead_code)]` attributes added here on `build_schema` and `schema_read_errors`, with the comment "Retained for callers outside the validation pipeline and focused compatibility tests."

Both functions are **live on current `main`**:
- `src/commands/mod.rs:282` — `schema::build_schema(&root.join(dir))`
- `src/commands/mod.rs:741` — `schema::schema_read_errors(&root.join(dir))`
- `src/mcp.rs:4264` — `crate::schema::build_schema(...)`

They're dead *on this branch* because this branch modifies `src/commands/mod.rs` (+23/−33) and **drops both call sites**. So the attribute isn't documenting an intentional API surface — it's silencing a warning that was correctly telling you the project-level surfacing of schema config problems and read errors had been removed.

That surfacing is the entire user-visible payoff of the #435 work. If it's gone, "unreadable migration files are flagged" is true of the function and false of the tool.

I suspect this is a stale-base artifact — the call sites arrived on `main` after `ce669736` — rather than a deliberate removal. Either way: please rebase, confirm both call sites are present afterward, and **remove the two `#[allow(dead_code)]` attributes**. If they're still needed after rebasing, that's the signal something is genuinely unwired.

**Also:**

1. **Conflicts with `main`** on `src/parser.rs`, `src/mcp.rs`, `src/commands/{check,comment,coverage,generate,issues,mod}.rs`, and eight spec files.
2. **Scope.** 13 source files, +2887/−431, closing one issue, while carrying whole-file copies of `parser.rs` and `validator.rs` that three other open PRs also touch. The description acknowledges this ("The whole current file is included here since #436 most owns both files"). I'd rather see `parser.rs` and `validator.rs` land as their own PR that the export-extraction and schema PRs can build on, than have four PRs racing on the same two files.

**On the substance** — the parsing fixes are right and several are security-relevant. Duplicate-key silent-override is the worst of them: a hidden second `status: draft` disabling all section and export checks is a validation bypass, and rejecting it loudly with the offending line is correct. Flow-style lists not parsing when `specsync new` scaffolds exactly that form is a bad interaction worth a regression test in the scaffold path too. Consolidating three divergent `depends_on` verdicts into one `validate_local_dependency` with path confinement is the right structural fix — that divergence, where some commands probed `..` escapes against the host filesystem and others ignored them, is a real vulnerability and not just an inconsistency.

---

---

0xLeif:

Issue #441, opened 2026-07-21
score undermines its own bar — JSON breaks on zero specs, explain arithmetic self-contradicts, untouched scaffold scores 85/B


0xLeif:
## Problem

`specsync score` undermines its own quality bar in four ways (verified on v5.2.0):

1. **JSON mode breaks exactly in the bootstrap case**: `score --format json` with zero specs prints plain human text (`No spec files found... Run specsync generate`) instead of JSON — breaking CI parsers on new projects; the no-match-filter path correctly emits `{"specs": []...}`.
2. **`--explain` arithmetic self-contradicts** in text and JSON — Sections criteria sum to 12/21 raw but the dimension reports 11/20 (the raw max of 21 is silently rescaled to 20); with `depends_on`, a spec's Freshness criteria sum 15/23 vs reported 12/20. Fails on all 56 specs tested.
3. **An untouched `specsync new` scaffold scores 85/B** — `placeholder_free` 6/6 and 4/7 "meaningful" sections despite being pure boilerplate (`Document this module's responsibility...`), so empty scaffolds already pass the documented ≥80 bar (see also #421's placeholder-generosity finding).
4. **Presence-only frontmatter checks**: `version: "notanumber"` → 5/5, `status: bogus` → 4/4, and vacuous `files_exist 15/15` when `files:` is missing entirely.

Also: `--explain` silently adds an `explain` key to the JSON schema, `--strict` exits 0 with F-grade specs, and empty-result JSON reports grade F with exit 0 (so the bar is unenforceable in CI either way).

PR #454, opened 2026-07-24
fix: generators emit self-valid specs; placeholder text scores as draft (closes #421)

### LINKED ISSUE #421: Generators (new/generate/scaffold/add-spec) emit specs that fail SpecSync's own validation; scoring hides it
## Problem

SpecSync's own generators produce specs that fail SpecSync's own validation, in three ways (all verified on v5.2.0):

1. **`add-spec` (and `scaffold` with no detected source file) writes `files:` as YAML null** — the very next `check --strict` exits 1 with `✗ Frontmatter invalid` on the file the tool just created.
2. **`generate`/`scaffold` auto-detect source files but leave Public API tables empty** — 0/N exports documented and a strict failure the moment the spec is promoted to `active`; `specsync new` pre-populates these correctly, so the generators disagree with each other.
3. **`specsync new` omits 3 of the config's own `required_sections`** (Invariants, Behavioral Examples, Error Cases) — promotion to `active` immediately fails strict check on the tool's own output.

Related scoring findings from the same probe, which make the self-validation gap worse by hiding it:

- A spec made of 100% the tool's own placeholder text scores **100/100 [A]** (placeholder sentences aren't recognized as draft markers), and an all-`TODO` spec scores exactly 80/100 [B] — the documented passing bar.
- The Freshness dimension never dropped below 20/20 in any test (stale 2020 changelog + freshly modified source).
- `specsync score` always exits 0 — even at F/40, even with `--strict` — so the documented ≥80 bar is unenforceable in CI.

## Suggested fix

Emit `files: []` (valid empty list) or detect-then-fill consistently across `new`/`generate`/`scaffold`/`add-spec`; treat the generator's own placeholder text as draft markers in scoring; add a `--min-score`/nonzero-exit mode so the quality bar is enforceable.

0xLeif:
## Problem

The generators emitted specs that failed the tool's own validation:

1. `generate_spec` wrote a bare `files:` (YAML null) when no source files were detected — an immediate `check --strict` failure (`Frontmatter missing required field: files`). `generate_spec_from_custom_template` already handled this correctly.
2. `generate`/`scaffold` left Public API tables empty (0/N exports documented, strict failure once promoted), while `new` pre-populated them.
3. Scoring hid all of it: a spec made of 100% generator placeholder text scored 100/A, and an all-TODO spec scored 80/B.

**Before** (v5.2.0, empty project): `specsync add-spec ghost` wrote
```
files:
  # - path/to/source/file
```
(a YAML comment → `files:` is null), and a spec built purely from the generators' placeholder sentences scored 100/A.

**After**: `files: []` is emitted, `scaffold auth` pre-populates `## Public API` with `| \`login\` | … |` rows from detected exports, and the same all-placeholder spec scores below A with draft-only-section suggestions.

## Root cause

- `generate_spec` unconditionally substituted `files:\n{files_yaml}\n` even when `files_yaml` was empty.
- No generator filled the Public API tables from detected exports outside `new`.
- `parser::STUB_PHRASES` didn't recognize the tool's own scaffold sentences, so `section_has_content`/`find_stub_sections` (and thus scoring's depth + A-grade cap) treated placeholder text as real documentation.

## Fix

- `generate_spec` emits `files: []` when the source list is empty (same as the custom-template path).
- New `collect_exports_for_files` + `populate_public_api_table` helpers insert a populated export table under `## Public API` whenever exports are detected (used by generate/scaffold/add-spec).
- The generator placeholder sentences (`Document this module's responsibility…`, `Define an invariant…`, `**Given** precondition`, etc.) are now stub phrases — sections containing only them count as draft-only, so the depth penalty and the ≥50%-stub A-grade cap apply. (Validation-side stub warnings are gated on non-draft so fresh scaffolds still pass `check`; that piece ships with #442.)
- Score exit-code semantics are intentionally untouched — that's W1's #441.

## Tests

`generate_spec_empty_files_emits_valid_empty_list`, `generate_spec_prepopulates_public_api_from_exports`, `populate_public_api_table_no_exports_is_noop`, `generator_placeholder_text_counts_as_stub`, `all_generator_placeholder_spec_does_not_score_a` — full suite green (1736 passed; the single failure is the pre-existing root-only `change::tests::non_git_walk_skips_volatile_trees_but_fails_on_relevant_walk_errors`), `cargo clippy -- -D warnings` green.

Note: `src/commands/new.rs` and `src/commands/scaffold.rs` changes for this issue (required sections in `new`, `add-spec` rewrite onto the shared generator) ship in the #442 PR since those files are owned by that workstream's changes — the bodies cross-reference.

Closes #421

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ specsync check --strict
 
 # Measure coverage and spec quality
 specsync coverage
-specsync score --all
+specsync score --all --min-score 80
 
 # Install native coding-agent workflows and Git hooks
 specsync agents install
--- a/site/src/content/docs/cli.md
+++ b/site/src/content/docs/cli.md
@@ -60,9 +60,11 @@ Quality-score your spec files on a 0–100 scale with per-spec improvement sugge
 ```bash
 specsync score                          # score all specs
 specsync score --json                   # machine-readable scores
+specsync score --min-score 80           # fail if any selected spec is below 80
+specsync score --strict                 # strict scoring implies min-score 80
 ```
 
-Scores are based on a weighted rubric: completeness, detail level, API table coverage, behavioral examples, and more.
+Scores are based on a weighted rubric: completeness, detail level, API table coverage, behavioral examples, and freshness. Scoring is advisory by default. `--min-score N` enforces a per-spec threshold from 0–100; `--strict` enforces at least 80. JSON adds `minimum_score` and `gate_passed` and remains parseable when the gate fails.
 
 ### `mcp`
 
@@ -453,6 +455,7 @@ specsync watch
 | `--dot` | Output dependency graph as Graphviz DOT (on `deps`). |
 | `--full` | Include companion files when creating a spec (on `new`). |
 | `--all` | Process all items, not just the first match (on `merge`, `score`). |
+| `--min-score N` | Fail when any selected spec scores below N (0–100, on `score`); `--strict` implies 80. |
 
 ---
 
@@ -461,7 +464,8 @@ specsync watch
 | Code | Meaning |
 |:-----|:--------|
 | `0` | All checks passed |
-| `1` | Errors found, warnings with `--strict`, or coverage below threshold |
+| `1` | Errors found, warnings with `--strict`, coverage below threshold, or a score gate failed |
+| `2` | Invalid CLI usage, including an out-of-range score threshold |
 
 ---
 
--- a/specs/cli/cli.spec.md
+++ b/specs/cli/cli.spec.md
@@ -1,11 +1,11 @@
 ---
 module: cli
-version: 10
+version: 11
 status: stable
 files:
   - src/main.rs
 db_tables: []
-tracks: [120]
+tracks: [120, 421]
 depends_on:
   - specs/agents/agents.spec.md
   - specs/archive/archive.spec.md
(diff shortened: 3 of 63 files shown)


0xGaspar (changes requested):
Requesting changes — rebase; plus one question about scoring that I'd like answered before this lands.

**Conflicts with `main`** on `src/validator.rs` and twenty-seven spec files. Note that `src/validator.rs` is also rewritten in full by #453 (+802/−53) and #458 (+92/−2); this branch's change to it needs to be reconciled with both, not resolved by picking a side.

**On the substance:** "the generators emitted specs that failed the tool's own validation" is the most embarrassing bug class in the batch and worth fixing exactly this way. `files:` emitting a YAML comment (hence null) rather than `[]` is a one-character-class bug with an immediate strict-mode failure, and the fact that `generate_spec_from_custom_template` already did it correctly makes it a clear oversight rather than a design question.

**The scoring change is the interesting part, and my one real question.**

Making the generator's own placeholder sentences count as stub phrases is clearly correct — a spec of 100% placeholder text scoring 100/A is a scoring function measuring the wrong thing entirely. But this changes scores for **existing** specs across every downstream project, with no code change on their side. A project sitting at A that drops to B or C on upgrade will read as a regression they caused.

- Is this called out as a breaking change in the changelog?
- Is there a way for a project to see *why* its score moved — i.e. does the score output name which sections were reclassified as draft-only?

I think the change is right and shouldn't be softened. It just needs to not surprise people.

**Also worth confirming:** the gating note says validation-side stub warnings are suppressed for drafts "so fresh scaffolds still pass `check`", and that this piece ships with #442 (#458). Since the scoring-side penalty ships **here** and the validation-side gating ships **there**, is there a merge order in which a fresh `new --full` scaffold both fails `check` and scores poorly? If so, please note the required order in both descriptions.

Rebase, answer the scoring question, and I'll approve.

---

---

0xLeif:

Issue #420, opened 2026-07-21
.specsyncignore per-spec syntax is dead; errors and the summary warning are unsuppressible


0xLeif:
## Problem

The documented per-spec `.specsyncignore` syntax (`<spec>:<pattern>`) is entirely dead — a dozen variants (`main:undocumented`, path keys, spacing/`=`/`@`/`#`/reversed separators) all silently no-op, while the same bare pattern works (verified on v5.2.0).

More of the matcher fails the common use cases:

- **Errors are unsuppressible** — no pattern touches phantom-export, missing-section, or missing-file errors.
- **The `⚠ N/M exports documented` summary warning is unsuppressible** by ~15 candidate IDs, so acknowledging an undocumented export can never yield a clean `--strict` run — arguably the most common intended use of the file.
- **Silent failure modes**: a UTF-8 BOM disables the first pattern, and one invalid-UTF-8 line poisons the whole file — both with zero diagnostics. Suppression itself is invisible in every output format (no "N suppressed" note; JSON `notices` stays empty), so typos and dead lines are indistinguishable from working rules.

The matcher itself, reverse-engineered: project-root `.specsyncignore` only, read by `check` only, matching internal kebab-case violation-kind IDs by case-insensitive dash-word-wise prefix (confirmed working: `undocumented`, `changelog`, `invariants`). Message text, paths, rule names, and globs never match.

## Suggested fix

Implement (or remove from docs) the per-spec syntax; make the exports-documented summary suppressible; emit one line of visibility (`N warning(s) suppressed by .specsyncignore`) and warn on unmatchable/invalid patterns; handle BOM and invalid UTF-8 explicitly.

PR #455, opened 2026-07-24
fix: db_tables validation silently passes dropped/renamed/misconfigured schemas (closes #435)

### LINKED ISSUE #435: db_tables silently unvalidated by default; schema_pattern vacuous; rename/drop inverted; quoting false-positives

0xLeif:
## Problem

`db_tables` validation silently passed in four situations:

1. **Dropped/renamed tables resurrected**: table discovery was a CREATE-pattern scan, so `DROP TABLE doomed` and `ALTER TABLE old_users RENAME TO users` left `doomed`/`old_users` in the discovered set — specs referencing dead tables passed.
2. **Quoting/case/schema prefixes**: `"Users"`, `` `users` ``, `USERS`, and `public.users` were treated as different tables.
3. **Missing `schema_dir` / uncompilable `schema_pattern`**: discovery returned an empty set and every declared table vacuously passed with no signal.
4. **Unreadable migration files**: silently skipped, vanishing tables and potentially no-op'ing all DB validation.

## Repro (fixture: `/var/tmp/scratch/w2/435/`)

```sql
-- migrations/001_init.sql
CREATE TABLE old_users (id INTEGER PRIMARY KEY);
CREATE TABLE doomed (id INTEGER);
-- migrations/002_migrate.sql
ALTER TABLE old_users RENAME TO users;
DROP TABLE doomed;
```
Spec declares `db_tables: [old_users, doomed, users]`.

- **Before**: all three pass — `old_users` and `doomed` are phantom-validated.
- **After**: `old_users` and `doomed` are "DB table not found in schema" errors; `users` passes. Quoted/cased/qualified spellings (`"USERS"`, `public.users`) match.

## Root cause

`get_schema_table_names` used a naive CREATE-only regex scan over migration files with no replay semantics and no name normalization, and every failure mode of the discovery path (missing dir, bad regex, unreadable file) was designed to fail open to an empty set.

## Fix

- `schema.rs`: migrations replay in filename order via `build_schema_with_retired` — `DROP TABLE` and `ALTER TABLE ... RENAME TO` retire names (returned alongside the table map); `normalize_table_name` strips per-segment quoting and lowercases; `schema_read_errors` flags unreadable migration files and unenumerable schema dirs.
- `validator.rs`: `get_schema_table_names` is replay-based (a `schema_pattern` scan can only ADD names, never resurrect retired ones); `schema_config_problems` (missing schema_dir / bad schema_pattern); `schema_table_exists` (normalized comparison, unqualified name matches schema-qualified discovered name); a `db_tables`-declaring spec with no configured `schema_dir` now warns visibly instead of skipping silently.
- `commands/mod.rs`: `run_validation` surfaces schema config problems and schema read errors once, project-level, as hard errors.

`test_get_schema_table_names_rename_and_drop_applied`, `test_schema_table_exists_quoting_case_and_schema_prefix`, `test_schema_config_problems_loud`, `test_validate_spec_db_tables_warn_when_no_schema_dir_and_error_on_phantom` (validator.rs); `test_quoted_and_schema_qualified_table_names`, `test_normalize_table_name`, `test_rename_and_drop_apply_to_quoted_names`, `test_schema_read_errors_flags_only_unreadable`, `test_schema_read_errors_flags_unenumerable_dir` (schema.rs).

## File overlap note

`src/validator.rs` (which contains the validator-side half of this fix) ships in PR #453 (`fix/436-frontmatter-validation`) — that issue most owns the file; the whole current file there includes these #435 changes. `src/commands/mod.rs` also contains the #420 "N warning(s) suppressed by .specsyncignore" visibility line, noted in the #420 PR. This PR ships `src/schema.rs` and `src/commands/mod.rs` in full.

Closes #435

### DIFF
--- a/specs/schema/context.md
+++ b/specs/schema/context.md
@@ -5,6 +5,10 @@ spec: schema.spec.md
 ## Key Decisions
 
 - **Migration replay ordering**: Files are sorted by filename, not modification time. This means migration files should use numeric prefixes (001_, 002_, etc.) for deterministic ordering.
+- **One fallible snapshot**: `build_schema_snapshot` owns directory loading, UTF-8 reads, canonical identity, and DDL replay. Validation consumers must use its `Result` instead of combining an empty map with a second best-effort error scan.
+- **Exact statement ordering**: Supported DDL is discovered outside SQL comments and replayed in byte order within each file. String literals and quoted identifiers do not create false operation boundaries.
+- **Collision semantics**: Plain duplicate CREATE, missing DROP/RENAME sources, and colliding RENAME targets fail without mutating the existing snapshot. IF EXISTS, IF NOT EXISTS, and OR REPLACE are the only explicit leniency/replacement paths.
+- **Canonical table identity**: ANSI double quotes, backticks, SQL Server brackets, case differences, whitespace around qualifiers, and qualified names pass through `canonicalize_table_name` for every CREATE/ALTER/DROP transition. Canonical rendering preserves dots inside quoted segments; `canonical_table_leaf` performs safe unqualified matching.
 - **Uppercase normalization**: Column types are always uppercased during SQL parsing to allow case-insensitive comparison with spec-documented types.
 - **Idempotent ALTER ADD**: If a column already exists in the table, `ALTER TABLE ADD COLUMN` is a no-op. This prevents duplicate columns from repeated migrations.
 - **Zero-dependency SQL parsing**: Uses regex-based parsing rather than a full SQL parser. Handles common DDL statements but not every SQL dialect edge case.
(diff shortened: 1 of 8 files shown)


0xGaspar (approved):
Approving the code. The description contains a claim that isn't accurate, which I'd like corrected before merge.

**Verified:**
- Merges **cleanly** into current `main`.
- Merged with #449/#456/#457: **2070 unit + 335 integration tests, 0 failures.**

**Description issue.** The Fix section says:

> `commands/mod.rs`: `run_validation` surfaces schema config problems and schema read errors once, project-level, as hard errors.

and the overlap note says "This PR ships `src/schema.rs` and `src/commands/mod.rs` in full."

Against the shared branch point this PR changes **only** `src/schema.rs` (+1336/−183) and `tests/integration/config.rs`. It does **not** touch `src/commands/mod.rs`.

I chased this because it looked like it would leave `schema_read_errors` orphaned. It doesn't — `main` already calls it at `src/commands/mod.rs:741`, and `schema::build_schema` at `:282`, so the wiring exists and the fix is genuinely complete. **No code change needed.** But the description implies this PR delivers wiring it doesn't contain, so please correct it.

**On the substance:** replay semantics are clearly right. A CREATE-pattern scan that never processes `DROP` or `RENAME` doesn't just miss things, it actively resurrects dead tables and phantom-validates specs against them — worse than no check. Making `build_schema_with_retired` return retired names alongside the table map, so a `schema_pattern` scan can only ever ADD and never resurrect, is the correct invariant and worth keeping as an explicit comment in the code.

Failing open to an empty set on missing dir / bad regex / unreadable file was the deeper bug — four independent paths to vacuous success. Good that all four now signal.

**One thing to watch:** #453 ships a near-identical `src/schema.rs` — the two differ by exactly two lines, both `#[allow(dead_code)]` attributes. See my review on #453; those attributes are load-bearing in a way that I think indicates a problem there, not here. Whichever lands second will need a real rebase, not a blind take-theirs.

Approving; please fix the description.

---

0xLeif:

PR #456, opened 2026-07-24
fix: .specsyncignore silently inert for BOM'd/corrupt/typo'd rules (closes #420)

### LINKED ISSUE #420: .specsyncignore per-spec syntax is dead; errors and the summary warning are unsuppressible

0xLeif:
## Problem

`.specsyncignore` had several silent-failure modes:

- A **UTF-8 BOM** silently disabled the first rule in the file.
- **One invalid-UTF-8 line** poisoned the whole file — every rule was dropped.
- Only one per-spec syntax order was accepted (`category:path`); the documented `path:category` order was ignored.
- **Typo'd/unmatchable rules did nothing** — indistinguishable from working suppression.
- The `N/M exports documented` partial-coverage summary warning had no category and could not be suppressed, so acknowledging undocumented exports could never yield a clean `--strict` run.
- Suppression itself was invisible: no way to tell how many warnings a rule swallowed.

## Repro (fixture: `/var/tmp/scratch/w2/420/`)

```
<BOOM>exports-documented
specs/ig/ig.spec.md: undocumented-export
undocumanted-export        # typo
```

- **Before**: BOM disables rule 1; the typo vanishes silently; the summary warning is unsuppressible either way.
- **After**: both valid rules apply (either syntax order), the typo'd line produces `.specsyncignore line 3 matches no warning category: 'undocumanted-export'`, the summary is suppressible via the new `exports-documented` category, and the check output ends with `ℹ N warning(s) suppressed by .specsyncignore`.

## Root cause

`IgnoreRules::load` read the file as one UTF-8 string (BOM and corrupt lines poisoned everything), accepted only one colon-order, and had no channel to report dead rules.

## Fix

- `ignore.rs`: BOM strip; per-line UTF-8 with a warning per skipped line; both per-spec orders (whichever side names a known category wins); unmatchable-line warnings; new `ExportsDocumented` category (classify `N/M exports documented` + `from_str("exports-documented")`).
- `commands/mod.rs`: surfaces `ignore_rules.warnings` in every output format and prints `N warning(s) suppressed by .specsyncignore`.

`test_load_tolerates_utf8_bom`, `test_load_invalid_utf8_line_skipped_not_poisoning`, `test_load_per_spec_path_first_order`, `test_load_warns_on_unmatchable_pattern`, `test_classify_exports_documented_summary`, `test_exports_documented_summary_suppressible`.

## File overlap note

`src/commands/mod.rs` ships in PR #455 (`fix/435-db-schema-validation`) — the whole current file there includes this visibility change. This PR ships `src/ignore.rs` in full.

Closes #420

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -283,6 +283,7 @@ pub fn run_validation(
     let mut all_errors: Vec<String> = Vec::new();
     let mut all_warnings: Vec<String> = Vec::new();
     let mut all_notices: Vec<String> = Vec::new();
+    let mut suppressed_warnings = 0usize;
     let mut file_owners: HashMap<String, Vec<String>> = HashMap::new();
     let mut spec_files_by_path: HashMap<PathBuf, HashSet<String>> = HashMap::new();
     for spec_file in ownership_spec_files {
@@ -352,12 +353,18 @@ pub fn run_validation(
             .map(|content| IgnoreRules::parse_inline(&content))
             .unwrap_or_default();
 
-        // Filter out suppressed warnings
-        let filtered_warnings: Vec<&String> = result
-            .warnings
-            .iter()
-            .filter(|w| !ignore_rules.is_suppressed(w, &result.spec_path, &inline_ignores))
-            .collect();
+        // Filter all configured suppressions, but attribute the visibility
+        // count only to project-file rules. Inline directives remain local to
+        // the spec and must not be reported as `.specsyncignore` activity.
+        let no_inline_ignores = HashSet::new();
+        let mut filtered_warnings: Vec<&String> = Vec::new();
+        for warning in &result.warnings {
+            if ignore_rules.is_suppressed(warning, &result.spec_path, &no_inline_ignores) {
+                suppressed_warnings += 1;
+            } else if !ignore_rules.is_suppressed(warning, &result.spec_path, &inline_ignores) {
+                filtered_warnings.push(warning);
+            }
+        }
 
         if collect {
             let prefix = &result.spec_path;
@@ -478,6 +485,10 @@ pub fn run_validation(
                     .map(|c| c.is_ascii_digit())
                     .unwrap_or(false)
         });
+        let suppressed_api_summary = result.export_summary.as_ref().is_some_and(|summary| {
+            result.warnings.iter().any(|warning| warning == summary)
+                && ignore_rules.is_suppressed(summary, &result.spec_path, &inline_ignores)
+        });
         if is_draft {
             println!(
                 "  {} Export validation skipped (status: draft)",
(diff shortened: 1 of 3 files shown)


0xGaspar (approved):
Approving the code — but the file-overlap note in the description is **backwards** and should be corrected before merge.

**Verified:**
- Merges **cleanly** into current `main`; merged with #449/#455/#457 the suite is green (2070 unit + 335 integration, 0 failures).

**The description says:**
> `src/commands/mod.rs` ships in PR #455 (`fix/435-db-schema-validation`) — the whole current file there includes this visibility change. This PR ships `src/ignore.rs` in full.

That's not what the branches contain. Against the shared branch point, **#455 does not touch `src/commands/mod.rs` at all** — its only source change is `src/schema.rs`. **This** PR is the one that ships `src/commands/mod.rs` (+40/−7), and it ships `src/ignore.rs` (+194/−9) as well.

The good news is that this makes #456 **more** self-contained than advertised, not less — nothing is missing. But the note as written would lead a reviewer (or a merge-order decision) to the wrong conclusion, so please fix it.

**On the substance:** every one of these is a real silent-failure class and the fixes are the right shape. Per-line UTF-8 decoding so one corrupt line can't poison the file, BOM stripping, and — most valuable — **warning on unmatchable rules**. A typo'd ignore rule that silently does nothing is indistinguishable from working suppression, which is the worst possible property for a suppression mechanism. The `N warning(s) suppressed by .specsyncignore` line is a good addition for the same reason: suppression should never be invisible.

Accepting either colon order (`category:path` and `path:category`) with "whichever side names a known category wins" is pragmatic. One edge case worth a test if it isn't covered: what happens when **both** sides name a valid category?

Approving; please correct the overlap note.

---

0xLeif:

PR #457, opened 2026-07-24
fix: export extraction misses/phantoms for TS unicode escapes, Erlang arity, GFM tables, NFC/NFD (closes #437, partial)

### LINKED ISSUE #437: Export extraction misses across backends — JS \u escapes, Unicode NFC/NFD, indented tables, Erlang/Lisp/Scala/C-header/Perl

0xLeif:
## Problem (partial fix — see Deferred below)

Export-extraction mismatches that produce phantom "no matching export" errors or silently hidden exports:

1. **TS/JS `\u` escapes**: `export const caf\u0061 = 4` (the legal identifier `cafa`) was truncated at the backslash to a bogus `caf` export — an unfixable phantom.
2. **Erlang arity**: exports were emitted as bare names, but Erlang identity is `name/arity` — `area/1` and `area/2` are different exports; one was always unmatchable and arity mismatches were invisible. Also `%` comments were only stripped on the file's last line (`(?m)` missing), so commented-out `-export` attributes on earlier lines leaked in.
3. **Indented GFM tables**: a 2-space-indented Public API table (valid CommonMark, renders on GitHub) was silently ignored by symbol extraction; duplicate rows were deduped silently.
4. **NFC vs NFD Unicode**: a spec documenting `naïve` (NFC) against source exporting `nai\u{0308}ve` (NFD) — glyph-identical — produced phantom errors.

## Repro (fixture: `/var/tmp/scratch/w2/437/`)

`src/geo.erl` (`-export([area/2]).`), `src/u.ts` (`export const caf\u00e9 = 1;`), spec with an indented GFM table documenting `area/2` and `café`.

- **Before**: `caf\u00e9` truncates to `caf`; Erlang export is bare `area`; indented table rows ignored.
- **After**: `café` decodes correctly; Erlang export is `area/2` (bare `area` in a spec still matches for back-compat, but `area/1` is correctly a phantom); indented tables parse; duplicate rows warn; NFC spec matches NFD source.

## Root cause / Fix

- `exports/typescript.rs`: `JS_IDENT` accepts `\uXXXX`/`\u{...}` sequences; `decode_js_identifier_escapes` decodes them (malformed input passes through untouched).
- `exports/erlang.rs`: emit `name/arity`; `(?m)` comment stripping; `-compile(export_all)` support (both atom and list forms).
- `parser.rs` (ships in PR #453): `TABLE_ROW_RE` tolerates 0–3 leading spaces; `get_duplicate_spec_symbols` warns on duplicate rows.
- `validator.rs` (ships in PR #453): `symbol_key` folds combining diacritics on both sides (with a dependency-free Latin decomposition table); `symbols_match` lets a bare spec name match an arity-qualified export (`area` ⇔ `area/2`) while `area/1` ≠ `area/2`.

`test_unicode_escape_identifiers_not_truncated` (typescript.rs); `test_erlang_commented_out_export_on_non_last_line_not_leaked`, `test_erlang_comment_with_dollar_dot_and_quotes_on_non_final_line`, `test_erlang_compile_export_all[_list_form]`, `test_erlang_export_list_and_callback_in_behaviour_module`, `test_erlang_inline_trailing_comment_inside_multiline_export_list` (erlang.rs); `test_get_spec_symbols_indented_table`, `test_get_duplicate_spec_symbols` (parser.rs, PR #453); `test_unicode_nfc_nfd_export_matching`, `test_erlang_arity_export_matching` (validator.rs, PR #453).

## Deferred (remaining #437 bullets, NOT in this PR)

- Lisp `defpackage`/`defgeneric`/`defclass` extraction
- Scala 3 `inline`/`given` extraction
- C header `typedef`s/`extern`s
- Perl `@EXPORT_OK` / case-insensitive matching

## File overlap note

`src/parser.rs` and `src/validator.rs` ship in PR #453 (`fix/436-frontmatter-validation`); the whole current files there include the #437 table/NFC-NFD/arity-matching changes above. This PR ships `src/exports/typescript.rs` and `src/exports/erlang.rs` in full.

Closes #437 (partial — deferred bullets listed above)

### DIFF
--- a/src/exports/erlang.rs
+++ b/src/exports/erlang.rs
@@ -12,7 +12,8 @@ static ERL_EXPORT_ATTR: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"(?s)-export\s*\(\s*\[\s*([^]]*)\s*\]\s*\)").unwrap());
 
 /// Function name within export list: name/arity
-static ERL_FUN_ARITY: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"'?\b(\w+)'?/\d+").unwrap());
+static ERL_FUN_ARITY: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"'?\b(\w+)'?/(\d+)").unwrap());
 
 /// `-compile(export_all).` or `-compile([export_all, ...]).`: a compiler
 /// directive that exports every top-level function in the module, regardless
@@ -39,8 +40,12 @@ pub fn extract_exports(content: &str) -> Vec<String> {
         if let Some(list_match) = caps.get(1) {
             let list_str = list_match.as_str();
             for f_caps in ERL_FUN_ARITY.captures_iter(list_str) {
-                if let Some(name) = f_caps.get(1) {
-                    let n = name.as_str().to_string();
+                if let (Some(name), Some(arity)) = (f_caps.get(1), f_caps.get(2)) {
+                    // Erlang export identity is name/arity: `area/1` and
+                    // `area/2` are different exports. Emitting the bare name
+                    // made one of them unmatchable (phantom) and hid arity
+                    // mismatches entirely.
+                    let n = format!("{}/{}", name.as_str(), arity.as_str());
                     if !symbols.contains(&n) {
                         symbols.push(n);
                     }
@@ -86,10 +91,10 @@ mul(A, B) -> A * B.
 helper() -> ok.
 "#;
         let symbols = extract_exports(src);
-        assert!(symbols.contains(&"add".to_string()));
-        assert!(symbols.contains(&"sub".to_string()));
-        assert!(symbols.contains(&"mul".to_string()));
-        assert!(symbols.contains(&"DummyClass".to_string()));
+        assert!(symbols.contains(&"add/2".to_string()));
+        assert!(symbols.contains(&"sub/2".to_string()));
+        assert!(symbols.contains(&"mul/2".to_string()));
+        assert!(symbols.contains(&"DummyClass/0".to_string()));
         assert!(!symbols.contains(&"helper".to_string()));
     }
 
@@ -109,8 +114,8 @@ real_fn(X) -> X.
 fake_fn(X) -> X.
 "#;
         let symbols = extract_exports(src);
-        assert!(symbols.contains(&"real_fn".to_string()));
-        assert!(!symbols.contains(&"fake_fn".to_string()));
+        assert!(symbols.contains(&"real_fn/1".to_string()));
+        assert!(!symbols.contains(&"fake_fn/1".to_string()));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


0xGaspar (approved):
Approving. Cleanest PR of the batch — +88/−23 across two files, fully self-contained.

**Verified:**
- Merges **cleanly** into current `main`.
- The file-overlap note is **accurate**: this branch touches only `src/exports/erlang.rs` and `src/exports/typescript.rs` vs the branch point. The `parser.rs` / `validator.rs` halves really do live in #453, as stated. This is the only PR in the batch whose overlap note I could confirm as correct.
- Merged with #449, #455, #456 and tested: **2070 unit + 335 integration, 0 failures.**

**On the substance:** treating Erlang export identity as `name/arity` is correct — `area/1` and `area/2` genuinely are different exports, and emitting bare names made one of them permanently unmatchable. Keeping bare `area` matching for back-compat while making `area/1` a true phantom is the right compromise. The missing `(?m)` on comment stripping is a nice catch; that class of bug silently leaks commented-out code into real output.

**Questions / non-blocking:**

1. `decode_js_identifier_escapes` passing malformed input through untouched — is that deliberate? Passing through means a malformed escape becomes a phantom export rather than a diagnostic. Reasonable either way, but worth a comment in the code saying which you chose and why.
2. The deferred bullets (Lisp, Scala 3 `given`/`inline`, C headers, Perl `@EXPORT_OK`) are clearly scoped out here — please make sure #437 stays open rather than being auto-closed, since the PR title says "partial". As written, `Closes #437 (partial …)` may still auto-close it on merge. Consider changing to `Refs #437` and closing manually.
3. `fmt` is red in CI; needs a `cargo fmt` pass.

Approving on the code. Point 2 is worth acting on before merge so the remaining work isn't lost.

---

0xLeif:

Issue #440, opened 2026-07-21
init/init-registry lie about detection and offer no repair path — fake source_dirs, TOML injection via --name, nested .specsync footgun


0xLeif:
## Problem

`specsync init` and `init-registry` lie about what they do, and offer no repair path (verified on v5.2.0):

1. **Fake source detection**: `init` prints `Detected source directories: src` and writes `source_dirs = ["src"]` even in a completely empty directory or one with only `app/`+`lib/` — no detection occurs; the message lies.
2. **No repair path**: re-`init` short-circuits on existing `config.toml` ("already exists", exit 0) and never restores deleted `sdd.json`/`version`/`.gitignore`, never diagnoses a corrupt config, and has no `--force`/`--repair` flag (clap even mis-suggests `--format`).
3. **`init-registry --name` is interpolated unescaped into TOML** — quotes/newlines produce a registry that then fails TOML parsing; `--name ""` is accepted; on an existing registry the command silently does nothing with exit 0.
4. **Subdir footgun**: running `init` inside a subdirectory of an initialized repo silently creates a nested `.specsync/` instead of detecting the parent project.
5. `--json`/`--format` is silently ignored by `init` and `init-registry`.

What held up: plain re-init is byte-identical idempotent, `.gitignore` is appended never clobbered, existing `specs/` untouched, non-git dirs work, and error quality for bad `--root`/blocking files is clean — zero destructive behavior.

PR #458, opened 2026-07-24
fix: new --full scaffold passes own check; enforce documented naming rules (closes #442)

### LINKED ISSUE #442: new --full produces a scaffold that fails the tool's own validation (7 scaffold-marker errors, greedy auto-detect)

0xLeif:
## Root cause

`specsync new --full` produced a scaffold that failed its own `check --strict`, and the scaffolding commands accepted module names the docs forbid:

1. **Missing required sections** — the `new` template omitted Invariants, Behavioral Examples, and Error Cases, so generated specs immediately failed frontmatter/section validation.
2. **Companion-marker warnings fired on drafts** — `validator.rs` emitted "Unfilled companion scaffold marker" warnings for `status: draft` specs (7 warnings on a fresh `new --full`), and stub-section warnings also fired on drafts; drafts are scaffolds by definition, so both checks are now gated on non-draft status.
3. **Greedy auto-detect** — add-spec/scaffold silently claimed an arbitrary single source file when no module-matching file existed; they now warn loudly when falling back.
4. **Name-validation gaps** — 200-char names, reserved names (`change`, `specs`, Windows device names like `con`), spaces/leading-dash/emoji, and case-fold collisions (`Lib` vs `lib`) were accepted. Added `validate_scaffold_module_name` (strict: 1-64 chars, documented charset, reserved list) used by new/add-spec/scaffold/wizard, while `validate_module_name` keeps the base traversal-safety rules for internal callers (e.g. change.rs specs named "change"). Added `check_case_collision`.
5. **Vague errors / undocumented rules** — frontmatter errors are reported inline with specifics, and `new --help` now documents the naming rules.

## Regression tests

- `commands::tests::validate_scaffold_module_name_rejects_overlong`
- `commands::tests::validate_scaffold_module_name_rejects_reserved`
- `commands::tests::validate_scaffold_module_name_rejects_spaces_leading_dash_emoji`
- `commands::tests::case_collision_detects_case_fold_duplicates`
- `validator::tests::draft_specs_skip_companion_scaffold_markers`

Verified end-to-end: `new --full` + `check --strict` exits 0 with 0 warnings.

## File overlap note

`src/cli.rs` also carries the `Init --repair` flag belonging to #440 (same file, disjoint change); the #440 PR includes the identical cli.rs content.

⚠️ **`src/validator.rs` is pending**: it exceeded the MCP push size limit during publication and is being pushed to this branch separately (blob sha `e996f84671e0723f71cbc44fe5b58c846e4a61cd`). The branch does not compile until it lands.

Closes #442

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -407,6 +407,7 @@ pub enum ChangeAction {
         /// Change ID
         id: String,
         /// Human actor authorizing the audited reopen transition
+        #[arg(long)]
         actor: String,
         /// Non-empty reason the accepted delivery evidence became stale
         #[arg(long)]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -90,7 +90,13 @@ pub enum Command {
         batch: Vec<String>,
     },
     /// Create .specsync/config.toml and initialize the verified SDD layout
-    Init,
+    Init {
+        /// Repair a partially-initialized project: restore missing .specsync
+        /// support files (version stamp, .gitignore, sdd.json, directories)
+        /// without touching the existing config
+        #[arg(long)]
+        repair: bool,
+    },
     /// Score spec quality (0-100) with letter grades and improvement suggestions
     Score {
         /// Show detailed per-category breakdown explaining exactly why each spec lost points
@@ -205,7 +211,10 @@ pub enum Command {
     },
     /// Quick-create a minimal spec for a module (auto-detects source files)
     New {
-        /// Module name for the new spec
+        /// Module name for the new spec.
+        /// Rules: 1-64 chars — letters, digits, `-`, `_`, `.`; must start with a
+        /// letter or digit; no spaces, path separators, or reserved names
+        /// (`change`, `specs`, Windows device names like `con`).
         name: String,
         /// Also create required companions (requirements.md, tasks.md, context.md, testing.md)
         /// and optional design.md when design artifacts are enabled
@@ -398,7 +407,6 @@ pub enum ChangeAction {
         /// Change ID
         id: String,
         /// Human actor authorizing the audited reopen transition
-        #[arg(long)]
         actor: String,
         /// Non-empty reason the accepted delivery evidence became stale
         #[arg(long)]
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -79,6 +79,29 @@ pub fn load_and_discover(root: &Path, allow_empty: bool) -> (types::SpecSyncConf
 /// misses. Control characters are rejected so a name cannot inject into the generated
 /// YAML frontmatter or create a control-char directory. Returns `Err` (to be printed and
 /// exited on) rather than writing outside the project.
+/// Maximum module-name length. Generous for real names while keeping the
+/// generated `<specs_dir>/<name>/<name>.spec.md` path far from filesystem
+/// limits (a 200+ character name otherwise dies with a raw OS error at mkdir).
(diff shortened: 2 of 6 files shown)


0xGaspar (changes requested):
Requesting changes — mostly rebase, plus a stale warning to clear.

**1. The pending-file warning appears to be resolved and should be removed.**

> ⚠️ `src/validator.rs` is pending: it exceeded the MCP push size limit during publication and is being pushed to this branch separately (blob sha `e996f84671e0723f71cbc44fe5b58c846e4a61cd`). The branch does not compile until it lands.

`src/validator.rs` at this head **is** `e996f84671e0723f71cbc44fe5b58c846e4a61cd` — it landed. Please strike that warning; right now it tells reviewers not to bother, which is presumably why this hasn't been looked at. Note `fmt` is still red in CI, so there may be a separate outstanding issue.

**2. Conflicts with `main`** on `src/commands/mod.rs` and `src/validator.rs`.

Worth flagging: this branch's `src/validator.rs` (+92/−2) and #453's (+802/−53) are different changes to the same file, and both PRs ship the whole file. These two need an explicit merge order and a real rebase, not a take-one-side resolution.

**3. `src/cli.rs`** — the description says this file's content is identical to #460's. It isn't (`e8496b95` here vs `de7c1edc` there). Minor, but it feeds into the merge-order question above.

**On the substance:** a `--full` scaffold that fails its own `check --strict` is the kind of bug that quietly teaches users to distrust the gate, so this is worth fixing properly. Gating companion-marker and stub-section warnings on non-draft status is the right call — drafts are scaffolds by definition and warning about it is noise.

The split between `validate_scaffold_module_name` (strict, user-facing) and `validate_module_name` (traversal safety, internal callers) is a good distinction. Two questions:

1. Reserved names including Windows device names (`con`, `aux`, `nul`, …) — is the check case-insensitive, and does it cover the `con.txt` form as well as bare `con`? Windows reserves those with any extension.
2. `check_case_collision` — is it Unicode-aware case folding or ASCII `to_lowercase`? Turkish dotless-i and similar will slip through an ASCII fold, which matters if this is meant as a collision guarantee rather than a heuristic.

Neither is blocking. Rebase and the stale warning are.

---

---

0xLeif:

PR #459, opened 2026-07-24
fix: exit codes & CI gating across report/coverage/deps/score/resolve/diff/changelog (closes #418, closes #419, closes #422, closes #425, closes #430, closes #431, closes #441, closes #444)

### LINKED ISSUE #418: changelog never validates refs — fabricates plausible output with exit 0 for unresolvable ranges
## Problem

`specsync changelog <range>` never validates that the range's refs resolve. Unresolvable refs are silently treated as an empty tree, producing plausible-but-fabricated changelogs with exit 0 — exactly the wrong output for CI.

Verified on v5.2.0 in a tag-less repo:

```
$ specsync changelog "v0.0.0-bogus..HEAD"
Spec Changelog: v0.0.0-bogus..HEAD
Summary: 1 added, 0 changed, 0 removed      # exit 0 — the ref does not exist

$ specsync changelog "HEAD~99..HEAD"         # beyond history in a shallow clone
# exit 0 with fabricated "added" entries as well
```

Two related failure modes from the same probe:

1. Three-dot ranges (`HEAD...HEAD~1`) are misparsed and report a phantom `1 removed` for a spec present in both commits — the opposite of reality, also exit 0.
2. `GITHUB_BASE_REF` auto-detection silently no-ops: bogus or valid values both print `No files changed since HEAD` with exit 0, so in GitHub Actions PR checkouts (shallow, detached) `specsync diff` passes without comparing anything. An explicit `--base bogus` errors properly, which makes the silent env path worse by contrast.

Minor: `specsync diff` outside a git repo answers with ~150 lines of raw `git diff` usage text as its error message.

## Suggested fix

Resolve both endpoints with `git rev-parse --verify` first and fail loudly (`unknown revision …`) before comparing; validate three-dot ranges against a real merge-base; apply the same validation to the `GITHUB_BASE_REF` path instead of silently treating it as HEAD.

### LINKED ISSUE #419: deps --require-coverage is inert; scalar depends_on silently disables the edge
## Problem

`specsync deps --require-coverage N` is completely inert: it exits 0 with no coverage message at any threshold — including `101`, and including `100` with unspecced source files present (verified on v5.2.0). The flag is documented as a coverage gate but enforces nothing.

Two more defects from the same probe:

1. **Scalar `depends_on: alpha` (a string, not a list) is silently accepted.** The edge vanishes from validation and graphing with no error from either `deps` or `check --strict` — a one-character frontmatter typo silently disables dependency enforcement.
2. **Duplicate `depends_on` entries aren't deduped** — they inflate the edge count and emit duplicate `alpha --> beta` lines in `--mermaid` output.

Also: `implements:`/`tracks:` frontmatter fields are neither validated nor graphed — any unknown frontmatter field passes `✓ Frontmatter valid` silently (worth deciding whether that's intentional).

## Suggested fix

Wire `--require-coverage` to the same coverage computation `specsync coverage` uses and fail below the threshold; reject (or normalize with a warning) scalar `depends_on`; dedupe edges at parse time.

### LINKED ISSUE #422: resolve --remote/--verify fetches the wrong registry path, prints false green, and always exits 0
## Problem

`specsync resolve --remote` and `resolve --verify` are end-to-end broken in three compounding ways (all verified on the published v5.2.0 asset):

**1. The fetch path doesn't match what the tool publishes.** `init-registry` and the docs publish the registry at `.specsync/registry.toml`, but the remote fetch requests `specsync-registry.toml` from the repo root — so every registry published by the tool 404s:

```
$ specsync resolve --remote
  ! Failed to fetch registry for CorvidLabs/spec-sync: HTTP request failed: http status: 404
  ? specs/lib/lib.spec.md -> CorvidLabs/spec-sync@cli (registry fetch failed)
```

(CorvidLabs/spec-sync has a committed `.specsync/registry.toml`; the docs' own "How It Works" section says the fetch targets `.specsync/registry.toml`.)

**2. False green on total failure.** `resolve --verify` with every fetch failing prints

```
  ✓ All cross-project references verified — no drift detected
```

**3. Exit code is always 0** — fetch failures, missing registries, network down, `--strict`, and the false-green verify all exit 0 (verified without pipelines). The documented CI usage (`specsync resolve --remote` in CI) therefore catches nothing, ever.

Related (same probe): `GITHUB_TOKEN` is never attached as an Authorization header despite the docs promising private-repo support, and `--json` is silently ignored.

## Suggested fix

Fetch `.specsync/registry.toml` (with the legacy root path as a fallback, not the primary), make fetch failures produce a drift/error verdict with exit 1 (and make `--strict` meaningful), never print "verified — no drift detected" when zero registries were successfully fetched, and attach `GITHUB_TOKEN` when set.

### LINKED ISSUE #425: coverage exits 0 while reporting 1 failed and prints a false all-referenced summary
## Problem

`specsync coverage` reports success on failure, twice over (verified on v5.2.0, exit codes measured without pipelines):

```
# active spec referencing a nonexistent file:
$ specsync coverage
  ✓ All source files referenced by specs          # false — src/missing.rs doesn't exist
1 specs checked: 0 passed, 0 warning(s), 1 failed
File coverage: 1/1 (100%)
exit code: 0
```

1. **Exit code is 0 with "1 failed"** — `coverage` never fails its process, unlike `check` (which correctly exits 1 on the same input, verified).
2. **`✓ All source files referenced by specs` prints even when a spec references a file that doesn't exist** — the ✗ `Source file not found` detail appears (in `check`), but the coverage summary claims the opposite.
3. **`--require-coverage 100` passes anyway** — missing files vanish from the denominator, so 1/1 is "100%" and the gate is vacuous (a related "vacuous 100%" guard exists elsewhere in the codebase but doesn't cover this path).

### LINKED ISSUE #430: report cannot gate CI — every enforcement path exits 0 on failure; commits_behind zeroed for non-stale modules
## Problem

`specsync report` cannot be used as a CI gate — every enforcement path exits 0 on failure (verified on v5.2.0, no pipelines):

- `--require-coverage 101` on a 100%-covered project: exit 0
- `--require-coverage 90` on a 0%-covered project: exit 0
- `--enforcement enforce-new` with unspecced files: exit 0
- `--strict` with stale + incomplete modules: exit 0

Additional defects from the same probe:

1. **`commits_behind` in `--format json` is zeroed unless the module is stale** — a module 6 commits behind reports `6` only once it crosses the stale threshold; below threshold it always reports `0`, hiding real sub-threshold drift from JSON consumers.
2. **`--format csv`, `table`, and `markdown` are silently ignored** (default text layout despite being listed in `--help`).
3. JSON modules lack a `status` field despite status-based filtering; filtered coverage mixes filtered numerator with unfiltered denominator; phantom source files (git-rm'd) stay listed with overall coverage at 100%.

### LINKED ISSUE #431: diff silently no-ops green in default CI states (detached HEAD, push events); pull_request_review missing from auto-detect
## Problem

`specsync diff` silently no-ops green in the default CI states it exists for (verified on v5.2.0):

1. **Detached HEAD with no env vars, or `GITHUB_EVENT_NAME=push`** (with or without `GITHUB_BASE_REF`): diff compares working tree vs `HEAD` — which is empty by construction in clean checkouts — and exits 0 with `No files changed since HEAD`. A push-event CI job always passes while comparing nothing, with no warning that auto-detection fell back.
2. **`GITHUB_EVENT_NAME=pull_request_review` is missing from the auto-detect allowlist** — GitHub sets `GITHUB_BASE_REF` for it, but specsync silently falls back to HEAD (no-op), while `pull_request` and `pull_request_target` work.

Related: an explicit `--base HEAD` is hijacked by the env auto-detect (HEAD doubles as the sentinel, so a literal HEAD comparison is impossible in PR context), and the `origin/` prefix is hardcoded (a remote named `upstream` fails loudly).

## Suggested fix

When auto-detection yields no usable base, say so (`no PR base detected; comparing against HEAD`) and offer `--require-base`/fail in strict mode; include `pull_request_review` in the allowlist; honor an explicit `--base` over env detection.

### LINKED ISSUE #441: score undermines its own bar — JSON breaks on zero specs, explain arithmetic self-contradicts, untouched scaffold scores 85/B

### LINKED ISSUE #444: resolve (local) exits 0 on unresolved deps; no path confinement; malformed refs vanish silently
## Problem

`specsync resolve` (local mode) shares the exit-code and validation gaps of its `--remote` sibling (#422), verified on v5.2.0 without pipelines:

1. **Exits 0 on unresolved local dependencies** — `✗`-marked unresolved refs print, but the process exits 0, and `--strict` doesn't change it (`check --strict` on the same project correctly exits 1).
2. **No path confinement on `depends_on`** — `/etc/passwd` and `../` traversal entries validate as ✓ in both `resolve` and `check --strict`; `deps` silently ignores the absolute path. Three commands give three different verdicts on the same dependency list.
3. **Malformed refs vanish silently** — entries like `repo@` (missing module) disappear from output with no diagnostic.
4. Minor: doubled heading "Local Local dependencies:"; `--json` silently ignored; blackhole network costs a 10s timeout per repo sequentially, misreported as `Connection refused`.

## Suggested fix

One verdict per dependency across check/deps/resolve: confine spec paths to the project, reject malformed refs with the offending text, and exit non-zero whenever any ref is unresolved (independent of `--strict`).

0xLeif:
## Summary
One consolidated PR for the W1 exit-code/CI-gating cluster (the changes share a common enforcement core — `commands::default_enforcement`/`compute_exit_code`/`exit_with_status`, `SpecSyncConfig.enforcement_set`, `CoverageReport.missing_files`, and main.rs wiring — so they land together).

**Core convention:** an explicit `enforcement` key in config is honored as-is (including `warn` → exit 0). When enforcement is UNSET, gate commands now default to gating on errors (validation failures exit non-zero) instead of the old implicit warn-everything. Wired into check, coverage, generate, comment, score, report.

## Issue → fix → regression tests
- **#418 changelog**: `parse_range_full` splits `...` before `..`; both endpoints validated via `git rev-parse --verify` (exit 1 on unknown revision); three-dot uses real `git merge-base`. Tests: `changelog_bogus_ref_exits_1`, `changelog_valid_range_exits_0`. Evidence: `changelog HEAD..bogus-ref` exit before=0 → after=1.
- **#419 deps**: `--require-coverage` wired to real coverage computation and gated; scalar `depends_on` normalized to a list with a warning; duplicate entries deduped at parse. Tests: `deps_scalar_depends_on_is_not_silently_dropped`, `deps_duplicate_depends_on_entries_deduped`, `deps_require_coverage_gate_enforced`. Evidence: scalar missing-dep exit 0→1.
- **#422 resolve --remote**: fetches `.specsync/registry.toml` first with legacy fallback; sends GITHUB_TOKEN/GH_TOKEN Bearer auth; fetch failures count as failed verification (exit 1); no false green when zero registries fetched; `--json` honored. Tests: covered by resolve suite + unit tests in registry.rs.
- **#425 coverage**: validation errors exit 1 by default; referenced-but-missing files counted in the coverage denominator and listed; false green suppressed; JSON gains `missing_files`. Tests: `coverage_missing_referenced_file_exits_nonzero_by_default`, `coverage_missing_referenced_file_warn_enforcement_exits_zero`, `coverage_json_includes_missing_files`. Evidence: spec referencing ghost file, `coverage` exit 0→1.
- **#430 report**: gating wired via compute_exit_code/exit_with_status (--strict/--enforcement/--require-coverage); stale+incomplete modules count as failures; `commits_behind` tracks sub-threshold drift; JSON status field; real CSV/Markdown renderers. Tests: `report_require_coverage_above_actual_exits_1`, `report_require_coverage_at_actual_exits_0`. Evidence: `report --require-coverage 101` exit 0→1.
- **#431 diff**: `pull_request_review` in the allowlist; `--base` is `Option<String>` so explicit beats env; HEAD fallback prints a loud stderr warning; `--strict` refuses the fallback (exit 1); git-diff failures exit 1 instead of silent "no drift". Tests: `diff_head_fallback_prints_loud_stderr_warning`, `diff_strict_refuses_head_fallback`. Evidence: before silent, after prints "specsync: no PR base detected … comparing against HEAD".
- **#441 score**: valid JSON for zero specs; Sections arithmetic made self-consistent; scaffold boilerplate counts as placeholder (`parser::is_boilerplate_line`); version/status values validated; files_exist non-vacuous; `--strict` exits 1 on F grades/zero specs. Tests: `score_json_with_zero_specs_outputs_valid_json`, `score_strict_with_zero_specs_exits_1`. Evidence: `score --json` on zero-spec project before=invalid JSON → after=valid JSON.
- **#444 resolve local**: unresolved local deps exit 1 regardless of --strict; `util::confine_path_to_root` rejects absolute/`..` paths; malformed `repo@` refs reported; "Local Local" heading fixed; one verdict/exit shared by text and JSON. Tests: `resolve_local_missing_path_exits_1_without_strict`, `resolve_local_outside_root_traversal_exits_1`, `resolve_malformed_cross_project_ref_exits_1`, `resolve_all_ok_exits_0`. Evidence: missing local dep exit 0→1.

Plus default-enforcement tests: `check_validation_errors_exit_nonzero_by_default`, `check_validation_errors_warn_enforcement_exits_zero`.

## Verification
- `cargo check --all-targets`: clean
- `cargo test`: 1720 unit tests pass (1 pre-existing root-only failure: `change::tests::non_git_walk_skips_volatile_trees_but_fails_on_relevant_walk_errors`, fails on main too); 223 integration tests pass; 20 new regression tests pass
- `cargo clippy -- -D warnings` (CI invocation): clean
- Before/after evidence generated with released binary vs this build

Closes #418, closes #419, closes #422, closes #425, closes #430, closes #431, closes #441, closes #444


0xGaspar (changes requested):
Requesting changes — CI is red and this is the largest merge-risk PR in the batch.

**1. CI failures.** `test (ubuntu-latest)`, `fmt`, `coverage`, `Packaged GitHub Action consumer` and `Required CI gate` are all failing; macOS and Windows were cancelled. The description reports a clean local run (1720 unit + 223 integration + 20 new). Please reconcile before re-review — with eight issues closing at once I can't tell which area is broken from the check names alone.

**2. This PR touches the most contended files in the repo.** 25 source files, including `parser.rs`, `validator.rs`, `registry.rs`, `config.rs`, `cli.rs` and `commands/mod.rs` — every one of which at least one other open PR also rewrites in full. It conflicts with `main` on `src/config.rs` and `src/validator.rs`, and it will conflict with #453, #458, #460 and #461 regardless of merge order.

I understand the rationale ("the changes share a common enforcement core … so they land together"), and I agree the `default_enforcement`/`compute_exit_code`/`exit_with_status` core should be one unit. But the eight issue-specific pieces mostly *consume* that core rather than define it. A first PR containing just the enforcement core plus its wiring, followed by the per-command changes, would be substantially easier to land and to revert if one of the eight turns out wrong.

**3. The behavioural change deserves prominent callout.** This is the important part and it's currently a sentence in the middle:

> When enforcement is UNSET, gate commands now default to gating on errors (validation failures exit non-zero) instead of the old implicit warn-everything.

That will turn currently-green CI red for every downstream project that never set `enforcement` — which is the correct behaviour and the entire point, but it's a breaking change to the tool's exit-code contract. It needs a changelog entry, a note in the release notes, and ideally a one-line migration hint in the error output telling users `enforcement = "warn"` restores the old behaviour. Honouring an explicit `warn` as-is is the right escape hatch; please make sure it's discoverable at the moment someone's build breaks.

**On the substance** the individual fixes look right, and several are the same "fail open = vacuous success" class the rest of this batch attacks: `--require-coverage` accepted but never wired to real computation, fetch failures counting as success in `resolve --remote`, `score --json` emitting invalid JSON for zero specs, silent "no drift" on git-diff failure. Splitting `...` before `..` in `parse_range_full` and validating both endpoints via `git rev-parse --verify` is correct — the old ordering would mis-split every three-dot range.

Happy to re-review once CI is green; the split is a strong suggestion, not a hard block if you'd rather keep it together.

---

---

0xLeif:

PR #461, opened 2026-07-24
fix: parse local registries with real TOML (closes #413)

### LINKED ISSUE #413: Local registry parser silently misparses documented [[modules]] format and accepts malformed TOML

0xLeif:
## Problem
Local registry handling (`src/registry.rs`) used a line scanner instead of a real TOML parser:
1. The documented array-of-tables modules format was silently misparsed — only mappings under the specs table were collected, so a registry written in the documented shape parsed as name = last module's name with zero mappings; every mapping was silently dropped.
2. Malformed TOML with a surviving `name=` line (e.g. `name = {{{`) was silently accepted, violating the fail-closed contract.

## Fix
- Replaced the line scanner with real TOML parsing (added `toml` crate dependency).
- Both the specs-table shape and the documented array-of-tables modules shape are supported; malformed module entries are rejected.
- Malformed TOML now hits the established `failed to parse local registry …` error; `is_inert_legacy_registry_stub` no longer treats garbage as inert.

## Regression tests
- `parse_registry_supports_documented_modules_shape`
- `load_local_registry_fails_closed_on_malformed_toml_with_name_line`
- `modules_entry_missing_spec_key_is_an_error`
- Previously failing `canonical_resolution_tolerates_inert_legacy_registry_stub` and both `semantic_application_*registry*` tests now pass.
- Drill reference: CorvidLabs/spec-sync-sandbox `drills/012-registry-parser-realities.sh`

## Validation
cargo test + `cargo clippy -- -D warnings` green (only pre-existing root-only failure `change::tests::non_git_walk_skips_volatile_trees_but_fails_on_relevant_walk_errors`).

Closes #413

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1097,6 +1097,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_spanned"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6662b5879511e06e8999a8a235d848113e942c9124f211511b16466ee2995f26"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
 name = "sha2"
 version = "0.10.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1148,6 +1157,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "tempfile",
+ "toml",
  "tree-sitter",
  "tree-sitter-c",
  "tree-sitter-commonlisp",
@@ -1290,6 +1300,45 @@ dependencies = [
  "displaydoc",
  "zerovec",
 ]
+
+[[package]]
+name = "toml"
+version = "1.1.3+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "53c96ecdfa941c8fc4fcaed14f99ada8ebed502eef533015095a07e3301d4c3c"
+dependencies = [
+ "indexmap",
+ "serde_core",
+ "serde_spanned",
+ "toml_datetime",
+ "toml_parser",
+ "toml_writer",
+ "winnow",
+]
+
+[[package]]
+name = "toml_datetime"
+version = "1.1.1+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3165f65f62e28e0115a00b2ebdd37eb6f3b641855f9d636d3cd4103767159ad7"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
+name = "toml_parser"
+version = "1.1.2+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2abe9b86193656635d2411dc43050282ca48aa31c2451210f4202550afb7526"
+dependencies = [
+ "winnow",
+]
+
+[[package]]
+name = "toml_writer"
+version = "1.1.2+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7d56353a2a665ad0f41a421187180aab746c8c325620617ad883a99a1cbe66d2"
 
 [[package]]
 name = "tree-sitter"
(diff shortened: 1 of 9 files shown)


0xGaspar (changes requested):
Requesting changes — small, and only really blocked on the rebase.

**1. Conflicts with `main` on `Cargo.lock`** (from adding the `toml` dependency against a stale base). Trivial to resolve by regenerating after rebase.

**2. `src/registry.rs` collides with #460.** #460's description claims both PRs ship an identical `registry.rs` (blob `04c19b38…`) so "either merge order is clean". That's not accurate — this PR's `registry.rs` is `826db07b8b4d97245edd55623d6ddbb2350a7b15`, #460's is `0cc2de76bdf7041710275f08fb6c8513d4f1bed6`, and neither matches the cited blob. Since this PR **replaces the line scanner with a real TOML parser** and #460 adds `toml_escape`/`generate_registry` on top of the old structure, whichever merges second needs a genuine hand-merge. I'd suggest landing this one first — parsing correctly is the foundation the escaping work sits on — and rebasing #460 onto it.

**On the substance this is my favourite fix in the batch.** Hand-rolled line scanning for a format with a real parser available is the root cause, and replacing it wholesale rather than patching the scanner is the right call. Two things I particularly like:

- The documented array-of-tables shape silently parsing as "name = last module's name, zero mappings" is a brutal failure mode — the tool's own documented format producing a silently empty result. Good catch.
- `is_inert_legacy_registry_stub` no longer treating garbage as inert. "Malformed TOML with a surviving `name=` line was silently accepted" is a fail-closed contract violation, and conflating "unparseable" with "legacy stub, ignore it" is how that happens.

The regression tests are well chosen, particularly `load_local_registry_fails_closed_on_malformed_toml_with_name_line` — it tests the contract rather than the implementation.

One question: with real TOML parsing, are there previously-"working" registries in the wild that were only working by accident of the line scanner's leniency? If so a migration note or a one-release deprecation warning may be kinder than a hard parse error. If the scanner was as broken as the description suggests, probably not — but worth a moment's thought.

Rebase and this is good to go.

---

---

0xLeif:

PR #463, opened 2026-07-24
fix: lifecycle performance cliffs in change check/list/correct-owner (closes #439)

### LINKED ISSUE #439: Lifecycle performance cliffs — change check superlinear at 20 changes; change list ~190ms/change; correct-owner batch ~4-8s/entry
## Problem

Lifecycle commands degrade superlinearly at modest scale (verified on v5.2.0):

1. **`change check` blows up at 20 changes**: 10 chained drafts 1.6s → 20 chained approved 15.8s → 20 with accepted/verifying evidence 20–52s. `sample` shows a `nanosleep` + `posix_spawnp` retry loop rather than real work, and the cost is driven by lifecycle *state* (20 chained drafts = 3.4s, same as unchained). Unusable as a CI gate at modest scale.
2. **`change list`/`change status` cost ~190ms per change** (~5s on 20 drafts, 9.3s on 50) due to per-change git subprocesses — slower than `change check` on the same data; `--format json` doesn't help.
3. **`change correct-owner` batch mode costs ~4–8s per applied entry** (full project rescan per entry, no progress output) — a 20-entry batch took 1m20s on a 21-file toy project. Fail-fast batches return in <1s, so the cost is per-applied-entry rescanning.

No correctness issues in the ordering/cycle logic at depth — purely performance cliffs in the three hottest lifecycle paths.

0xLeif:
## ⚠️ Status note
This PR currently contains 1 of 2 changed files. **`src/change.rs` (860KB) could not be uploaded through the automation channel** — the complete 31KB patch (`w6-439.patch`, applies cleanly on main @ ce66973, commit 73ad0d1) is attached to the task report. The branch will not compile until the change.rs portion lands.

## Problem
Lifecycle commands degraded superlinearly at modest scale: `change check` hit double-digit seconds at 20 changes, `change list`/`status` cost ~190 ms per change, and `change correct-owner` batches revalidated every correction after each appended entry (O(entries^2) Git-backed ownership lookups).

## Root causes
- Read-only lifecycle surfaces repeated repository-global Git lookups and the full (doubled) Git evidence pipeline for every change, and terminal-evidence validation recursed over accepted chains with a fresh memo per change — thousands of git subprocess spawns at 20–50 changes.
- correct-owner batches re-validated all pre-existing corrections after every appended entry, each with its own canonical-spec evidence collection.

## Fix
- Read-only commands (list/status/show/check) now run in a scoped per-command memoization scope: repository-global Git facts, per-candidate-set evidence, project input digest, and terminal-evidence validity are computed once per command. Mutating commands never install the scope; the Git index race guard still re-reads index contents every collection.
- correct-owner validates pre-existing corrections once up front and resolves canonical module ownership at most once per module (fail-fast entry naming and error messages unchanged).
- The bounded Git runner polls on a 200µs quantum instead of 2ms.

## Measured (debug builds, 20/50 chained changes, half accepted)
| Command | Before | After |
|---|---|---|
| change check @20 | 11.2s (3621 git spawns) | 1.4s (599 spawns) |
| change list @20 | 13.0s | 1.9s |
| change list @50 | 30.6s | 3.8s |
| change check @50 | 16.0s | 4.1s |
| correct-owner 20-entry batch | 19.9s | 0.25s |

Scaling is now near-linear. Regression tests (`read_scope_memoizes_repeated_summary_git_lookups`, `read_scope_reuses_git_evidence_for_repeated_candidates`) assert identical summaries with strictly fewer git subprocess spawns (spawn-count hooks, no wall-clock timing). Full `cargo test` green (except one pre-existing root-only failure) and `cargo clippy -- -D warnings` green.

Closes #439

### DIFF
--- a/specs/change/change.spec.md
+++ b/specs/change/change.spec.md
@@ -1,6 +1,6 @@
 ---
 module: change
-version: 43
+version: 45
 status: active
 files:
   - src/change.rs
@@ -35,6 +35,8 @@ Provides the spec-sync 5.0 verified spec-driven development lifecycle, including
 15. Supported accepted interview metadata changes only through a portable append-only correction ledger whose effective definition requires fresh gates and never replays canonical deltas.
 16. Audited exact acceptance-owner corrections can repair omitted canonical ownership on an already-scoped input without changing semantic scope or replaying canonical deltas.
 17. A transactional batch of audited exact acceptance-owner corrections validates every entry independently and persists all or none as sequenced ledger entries.
+18. Read-only lifecycle commands and unified check entry points use one bounded invocation snapshot for stable repository facts, evidence, records, and terminal validity; mutation APIs remain live and uncached.
+19. Dependency ordering is deterministic regardless of discovery order, and owner-correction batches resolve canonical ownership once per module.
 
 ## Public API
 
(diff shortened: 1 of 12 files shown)


0xLeif:

PR #462, opened 2026-07-24
fix: lifecycle state machine hardening (closes #411, closes #412, closes #424, closes #426, closes #434, closes #443, closes #445, closes #432, closes #433, closes #423)

0xLeif:
## ⚠️ Status note
This PR currently contains 11 of 12 changed files. **`src/change.rs` (860KB) could not be uploaded through the automation channel** and must be added to this branch (expected git blob SHA `7252c4190a306683e3fdbba4116ae4f1312a652b`; full file and a 71KB patch vs main are attached to the task report / available from the author). The branch will not compile until it lands. Everything else is byte-verified against the validated local commit.

## Summary
Consolidated PR for the W3 lifecycle workstream — the fixes interleave inside the same lifecycle files (src/change.rs, src/commands/change.rs), so they land together. Validation at the full local HEAD: 1734 unit + 228 integration tests pass, `cargo clippy -- -D warnings` clean, only failure is the pre-existing root-only `change::tests::non_git_walk_skips_volatile_trees_but_fails_on_relevant_walk_errors` (fails on main too when run as root).

## Issue → fix → tests
- **#411 cryptic no-remote error**: `uncovered_meaningful_paths` no longer flattens git stderr — new `git_output_allow_empty_result` preserves it; error names cause + remediation ("SDD coverage needs a comparison base: no remote default ref found and no recorded base commit; add a remote … or run `specsync change adopt`"). Tests: `missing_comparison_base_error_names_cause_and_remediation`, `unreachable_recorded_base_error_names_the_base`.
- **#412 stray archive dir**: stray dirs under .specsync/archive/changes (no state.json, no files) skipped with a stderr Warning naming path + remediation; entries containing ledger files still hard-fail (tampered archives never silently skipped). Test: `stray_archive_directory_without_state_is_skipped_with_warning`.
- **#424 approve on verifying**: re-approval of a verifying change now prints a loud `reapproval_discards_verification_warning` naming the discarded evidence + re-verify remediation (refusal impossible: 12 existing tests assert the transition itself). Test: `reapproving_verifying_change_warns_that_verification_is_discarded`.
[...]

PR #483, opened 2026-07-31
fix: combine open pre-6.0 fix PRs onto SpecSync 6.0

### LINKED ISSUE #413: Local registry parser silently misparses documented [[modules]] format and accepts malformed TOML

### LINKED ISSUE #418: changelog never validates refs — fabricates plausible output with exit 0 for unresolvable ranges

### LINKED ISSUE #419: deps --require-coverage is inert; scalar depends_on silently disables the edge

### LINKED ISSUE #420: .specsyncignore per-spec syntax is dead; errors and the summary warning are unsuppressible

### LINKED ISSUE #421: Generators (new/generate/scaffold/add-spec) emit specs that fail SpecSync's own validation; scoring hides it

### LINKED ISSUE #422: resolve --remote/--verify fetches the wrong registry path, prints false green, and always exits 0

### LINKED ISSUE #425: coverage exits 0 while reporting 1 failed and prints a false all-referenced summary

### LINKED ISSUE #430: report cannot gate CI — every enforcement path exits 0 on failure; commits_behind zeroed for non-stale modules

### LINKED ISSUE #431: diff silently no-ops green in default CI states (detached HEAD, push events); pull_request_review missing from auto-detect

### LINKED ISSUE #436: Frontmatter parser leniency is a validation bypass — duplicate keys last-wins; depends_on entries never validated; no path confinement

0xLeif:
## Summary

One PR instead of ~15 separate CI matrices. Rebases/merges the open pre-6.0 fix branches onto current `main` (SpecSync 6.0 / #480) with conflict resolution that **keeps 6.0 lifecycle semantics**.

### Strategy

- Merged each open fix branch with `-X ours` so 6.0 wins conflicts.
- Restored `src/change.rs` (and related lifecycle adapters) to **6.0 main** — pre-6.0 lifecycle PRs re-pasted entire read-scope / lock stacks and would not compile cleanly on 6.0.
- Deduped merge artifacts in hooks / typescript / command modules.
- Re-stitched non-lifecycle helpers that still apply (naming rules, frontmatter guards, shared dependency verdict, default enforcement helpers).

### Supersedes (close these PRs after merge)

- #453 frontmatter validation
- #454 generator self-valid specs
- #455 db schema validation
- #456 .specsyncignore
- #457 export extraction
- #458 new/scaffold naming
- #459 exit codes & CI gating
- #460 init honesty / repair
- #461 registry real TOML
- #462 lifecycle hardening *(code largely not ported — 6.0 redesign owns this; leave issues open if still valid)*
- #463 lifecycle perf *(same)*
- #471 reopen manifest growth *(same — re-evaluate on 6.0)*

### Issues

Non-lifecycle (intended to land here):

- Closes #413
- Closes #418
- Closes #419
- Closes #420
- Closes #421
- Closes #422
- Closes #425
- Closes #430
- Closes #431
- Closes #436
- Closes #437
- Closes #440
- Closes #441
- Closes #442
- Closes #444

Lifecycle issues from #462/#463/#471 are **not** auto-closed — 6.0 already rewrote that surface; file follow-ups if any gap remains after dogfood.

Partial / verify-on-CI:

- #435 schema validation (schema module kept 6.0 main; confirm coverage in CI)

### Test plan

- [x] `cargo check` green on combine branch
- [ ] `cargo test` (running locally + CI)
- [ ] CI Required gate / trust / archive as applicable
- [ ] After merge: close superseded PRs #453–#463, #471


0xLeif:
**Status after main hygiene (#484) + docs (#482) landed:**

- Main now has CHG-0068/0069 **finalized & archived** and docs refresh merged.
- This combine PR still has **product test failures** (warm-cache, coverage, import, etc. from half-merged pre-6.0 fix PRs) — **not** just lifecycle staleness.
- Recommend **not** merging as-is. Next step: slim to green non-lifecycle hunks or re-port fixes one-by-one on current main.

Leaving open for rework rather than force-merge.

Issue #533, opened 2026-08-11
bug(change): `change check --commit` self-commits change-sequence.json backwards, regressing the high-water mark


0xLeif:
**Version:** specsync 6.0.0 (product `main` @ `f3a65dc5`). Observed live on this repo while shipping CHG-0101; then reproduced from scratch in a throwaway repo.

## What happens

1. `change new` writes the newly minted sequence into `.specsync/change-sequence.json` **in the working tree only** (`src/change.rs:1528` → `update_change_sequence_claim`, `src/change.rs:1660`). It never commits it, and the `.specsync/changes/CHG-N/` workspace stays untracked. It also force-adds the ledger to the change's `affected_paths` (`src/change.rs:1466`), so the change owns the file as a delivery input.
2. Shared `main` advances past `N`. The branch is brought fully up to date (0 commits behind `origin/main`), so `HEAD` carries main's higher ledger while the author's older uncommitted copy still claims `N`.
3. `change check --commit` calls `git_commit_all` = `git add -A` + `git commit` (`src/commands/change.rs:2252`, `2291-2302`) with **no high-water validation**, and its own `chore(lifecycle): materialize <id>` commit records the ledger backwards.

Live evidence — commit `6576162d` on `0xleif/finalize-chg0101`, parent `f3a65dc5` (= `origin/main`):

```diff
--- a/.specsync/change-sequence.json
+++ b/.specsync/change-sequence.json
 {
    "schema_version": 1,
-  "sequence": 103,
-  "id": "CHG-0103-address-pr-531-review-by-validating-correction-ledger-health-before-mutating-lif",
+  "sequence": 101,
+  "id": "CHG-0101-add-audited-solo-maintainer-self-review-override",
```

That same commit adds the entire `CHG-0101` workspace as new files (`state.json`, `approvals.json`, `verification.json`, `verification-attempts.json`), confirming nothing had been committed before it. `CHG-0101/state.json` records `base_commit f1e7d3ab` — a main whose ledger read `100` — and a `created_at` three days earlier.

## Reproduction

Build-free, ~100s, 18 assertions: `CorvidLabs/spec-sync-sandbox` PR #46, `drills/037-sequence-ledger-regression.sh`. Observed there:

```
$ specsync change check CHG-0002-stale-branch-change --commit
Checking (1/2): verifying the materialized tree…
Committed the materialized spec.
Checking (2/2): re-verifying against the committed tree…
Committed the verification evidence.
[...]

Issue #537, opened 2026-08-11
bug(change): the semantic delta format is undiscoverable, and a populated file reports "is empty"


0xLeif:
**Version:** specsync 6.0.0 (`main` @ `f3a65dc5`). Found by a cold-agent UX audit driven using only the tool's own output.

## Summary

The semantic delta is a hard gate on `change approve`, and its format is documented nowhere reachable by a user or an agent. The word "delta" appears in the entire cold-agent surface exactly once — in `change check`'s one-line description. It is absent from `--help`, from every error message, and from the generated `SKILL.md`.

The auditing agent burned roughly 25 failed commands brute-forcing the grammar, then had to read `src/change.rs:7960-7975`. That was the **only** point in the whole audit that required reading the product's source.

## The undocumented contract

- File at `.specsync/changes/<id>/deltas/<module>.md`
- `## Added` | `## Modified` | `## Removed`
- then `### REQUIREMENT <key>` or `### SPEC SECTION <key>`

## Two defects in the parser's own behaviour

**1. Case sensitivity is inconsistent between the two heading levels.** `##` is case-insensitive (`to_ascii_uppercase`); `###` is case-sensitive (raw `strip_prefix("REQUIREMENT ")`). So `## added` is accepted while `### requirement REQ-greeter-001` is not:

```
$ printf '## Added\n\n### requirement REQ-greeter-001\n\n...' > deltas/greeter.md
error: invalid delta item heading `### requirement REQ-greeter-001`
```

**2. A populated file is reported as empty.** Any file whose headings are unrecognized produces "is empty", which sends the author looking for a missing file rather than a malformed one:

```
$ printf '# greeter\n\nAdds greeting format docs.\n' > deltas/greeter.md
error: semantic delta for `greeter` is empty          # the file has three lines
```

## Impact

`approve` cannot be completed without this knowledge, so it blocks the guided path at its first real gate. An agent driving from `--help` cannot construct a valid delta at all.

Note the contrast: the tool's *other* delta error messages are among the best in the product — `use ## MODIFIED to change a block already present in the living tree` and `add a row for it to the '## Requirement evidence' table in <path>` are both exemplary. The grammar simply has no entry point.

## Suggested fix

1. Enumerate the allowed values in the error: `expected one of: Added, Modified, Removed`, and likewise for the `###` level.
2. Make `###` case-insensitive, matching `##`.
3. Replace `is empty` with `contains no recognized delta operation headings` when the file has content.
[...]

Issue #536, opened 2026-08-11
bug(change): ship writes an empty archive deltas/ dir; the orphan breaks every command after a branch switch while git reports clean


0xLeif:
**Version:** specsync 6.0.0 (`main` @ `f3a65dc5`). Found by a cold-agent UX audit.

## Summary

`change ship` writes an empty `deltas/` directory into the dated archive package. Git cannot track an empty directory, so it survives any checkout of a commit that predates the archive — as an untracked orphan that `git status` reports as clean. Every subsequent spec-sync command then hard-fails on the missing `state.json`.

## Reproduction

```
$ specsync change ship CHG-0001-bootstrap                        # exit 0
$ git add -A && git commit -m archive
$ git ls-files .specsync/archive | wc -l                     ->  15
$ git ls-files .specsync/archive/changes/*/deltas | wc -l     ->   0    # empty dir, untrackable
$ git checkout <older-commit>
$ git status --short                                          -> (clean)
$ find .specsync/archive
.specsync/archive/changes/2026-08-11-CHG-0001-bootstrap/deltas          # orphan survives
$ specsync check
error: failed to read archived change state /.../2026-08-11-CHG-0001-bootstrap/state.json: No such file or directory (os error 2)
EXIT=1
```

`specsync check`, `change audit`, and `change new` all die the same way. Recovery is `rm -rf` of the orphan directory, which is stated nowhere in the tool's output.

## Impact

This fires on every ordinary git operation that moves to a pre-archive ref: `git checkout main`, `git bisect`, branch switching, and any CI job that checks out an older commit. `git status` reporting a clean tree makes the cause genuinely hard to find — the failure names a file the user never created in a directory git says is not there.

Related: this is the same shape as issue #412 (stray directory under `.specsync/archive/changes` hard-fails ledger enumeration), but reached through the tool's own normal output rather than by hand.

## Suggested fix

Both halves are worth doing:

1. **Tolerate it at read time** — skip archive directories with no `state.json` rather than erroring, or report them as a warning naming the remedy. Enumeration should not be brittle to a stray directory (see #412).
2. **Stop creating it** — do not write an empty `deltas/` at all when the change has no deltas, or write a `.gitkeep` so git can track it and checkout removes it cleanly.

0xLeif:
## Re-measured on `446aacd2` — five claims, one of them shifted

[...]

Issue #535, opened 2026-08-11
bug(change): the whole lifecycle completes with zero commits and leaves an unrecoverable archived repo


0xLeif:
**Version:** specsync 6.0.0 (`main` @ `f3a65dc5`). Found by a cold-agent UX audit driven using only the tool's own output.

## Summary

Nothing in the lifecycle requires a commit. `change check`, `review`, and `ship` all succeed against a tree where the change's work is entirely untracked — and the resulting archived state cannot be committed by any means, with **every recovery verb refusing**.

## Reproduction

`change show`'s `next_action` says plain `run specsync change check <id>` — it never mentions `--commit`. Following it literally:

```
$ specsync change check CHG-0001-...                      -> ✓ verified                      (exit 0)
$ specsync change review CHG-0001-... --reviewer other-human
  ✓ ... independent review recorded as pass at e70054c3122763b74f53388839e0d0e5d62f47ea      (exit 0)
$ specsync change ship CHG-0001-...
  ✓ CHG-0001-... finalized on this PR                                                        (exit 0)
```

`e70054c3` is the repository's **initial commit**. `git log` shows one commit; the entire change was untracked the whole time.

Then, obeying `ship`'s own closing instruction (`Next: commit if needed`):

```
$ git add -A && git commit -m "Add: greeter documentation module (CHG-0001)"
error: CHG-0001-...: archived change historical integrity is invalid: accepted change
  `CHG-0001-...` requires exactly one trusted transition matching its state,
  verification, and closing evidence; found 0
error: unable to inspect changed paths for SDD coverage
EXIT=1
```

## No recovery path exists

`change list` reports `No active SDD changes.` Every recovery verb refuses:

```
change check         -> cannot check the change while ... is archived
change reopen        -> accepted change delivery inputs are current (exact or successor-covered);
                        reopen is allowed only when delivery evidence is stale
change correct       -> cannot correct accepted interview metadata while ... is archived; expected accepted
change correct-owner -> cannot correct an acceptance input owner while ... is archived; expected verifying
change review        -> cannot record scoped review while ... is archived; expected verifying
change finalize      -> cannot finalize the change while ... is archived; expected verifying or accepted
```

Note that `reopen`'s refusal ("delivery inputs are current") **directly contradicts** the commit hook ("historical integrity is invalid"). The only escape is `git commit --no-verify`, after which `specsync check` still exits 1.

## Impact

[...]

PR #538, opened 2026-08-11
feat(change): audited solo-maintainer self-review override (CHG-0101)

### LINKED ISSUE #530: bug(change): concurrent check --commit returns success with stale verification evidence
## Summary

Two concurrent `specsync change check <id> --commit` commands can both exit 0 while leaving both active changes with verification evidence that `specsync change audit --strict` immediately rejects as stale. A zero exit must not claim verified state that the next lifecycle gate rejects.

## Reproduction

Run the WIP 6 binary against the sandbox drill:

```bash
RESULTS_DIR=/tmp/specsync-sandbox-results \
SPECSYNC=/path/to/specsync \
./drills/035-concurrent-check-serialization.sh initial
```

The drill creates a throwaway Markdown-only repository (no language build), initializes `verification_commands: ["true"]`, creates and approves two independent `--no-spec-change` documentation changes, modifies distinct files, then starts both `change check --commit` calls concurrently.

## Observed on SpecSync 6.0.0 WIP

- both first checks exit `0`;
- both workspaces contain parseable `verification.json`;
- `change audit --strict` exits non-zero for **both** IDs with `verification evidence is stale for the current commit or contract`.

This is an unsafe success result around the shared commit/tip stage. The command should serialize correctly or fail/retry explicitly; it must not return success with stale evidence.

## Regression coverage

Sandbox PR will add drill 035. It currently passes only when this failure is reproduced, and will invert once this issue is fixed.

0xLeif:
## Summary

Adds an explicit, audited solo-maintainer exception to the scoped review gate:

```
specsync change review <id> --self-review --actor <scope-approver> --reason "…"
```

- Accepted **only** when `--self-review`, `--actor`, and a non-empty `--reason` are all supplied and the actor matches the approved scope approver.
- Persists the exception, actor, and reason in durable review evidence (`ScopedReviewMode::SelfReview`, `required_check: "self_review"`), and marks the status as self-reviewed.
- Independent `--reviewer` remains the default and is unchanged; malformed or mismatched identities are rejected.
- Does **not** bypass verification, trust, CI, or finalize.

Requirements: `REQ-change-046`, `REQ-cmd-change-005`, `REQ-cli-args-009`.

## Test Plan

- [x] `cargo test` — full suite green (334 integration passed, 0 failed, in 291s), run through `change check --commit` so evidence binds the committed tree
- [x] `python3 -S .github/scripts/validate-release-version.py` — `Validated release version 6.0.0 across current distribution surfaces`
- [x] `python3 -S .github/scripts/validate-workflow-runtime-pins.py` — `Validated oven-sh/setup-bun@v2 with Bun 1.3.14 across 3 workflow jobs`
- [x] `fledge lanes run pre-push` — 62/62 specs, 105/105 files, 128120/128120 LOC, PASS in 73s
- [x] Reviewed via the feature itself (`--self-review`), with a rebuilt binary, after the product tip was green

## Note on the extra lifecycle commits

This branch carries two `materialize` + `record verification` pairs rather than one. The first pair committed `.specsync/change-sequence.json` **backwards** — 103 → 101 — clobbering the high-water mark `main` already carried. The pre-push gate caught it:

```
error: change sequence ledger claims CHG-0101 but the highest recorded sequence is CHG-0103
```

The ledger was restored from `origin/main` and re-anchored with a second `change check --commit`. Root cause is filed as #533: `check --commit` uses `git add -A` with no pathspec, so it commits the working tree rather than the change's declared paths. That also explains the cross-change contamination noted in #533's comments and the drill self-contamination that closed #530.


0xLeif:
Cancelling this change rather than landing it.

CHG-0101 added an audited solo-maintainer self-review override — an override for the separation-of-duties rule. The direction has since changed to removing that rule entirely: approval, scoped review, commit binding, and archive integrity are being delegated to `attest`, which already records exactly this (signed, keyed to commit SHAs, in git notes) and does it in a storage layer that survives history rewrites. An override for a rule that is being deleted is scaffolding.

Since 6.0 is unreleased — latest tag is `v5.2.0` — the reduction lands before the tag rather than as a breaking change afterwards.

The branch is left in place; nothing is lost. What this change taught is recorded rather than discarded, which is also why cancellation is now filed as a first-class lifecycle state (#541): today the only way to abandon a change is `rm -rf`, and the knowledge goes with it.

Findings that came out of this change and survive it: #533 (unscoped staging in `check --commit`), #534, #535, #536, #537 (cold-agent UX audit).

PR #556, opened 2026-08-13
fix(6.0): stop printing green result lines for checks that could not run (#553)

### LINKED ISSUE #553: A spec with no frontmatter still reports `✓ All required sections present` when six are missing
When frontmatter fails to parse, validation stops gathering configuration but still prints affirmative results for checks that either did not run or are false.

## Repro

`specs/nofm/nofm.spec.md` containing only `## Purpose` and `## Public API`, and no `---` block at all:

```
  ✗ Frontmatter invalid
    ✗ Missing or malformed YAML frontmatter (expected --- delimiters)
  ✓ All source files exist
  ✓ All required sections present
  ✓ All dependency specs exist
exit 1
```

Six of the eight default required sections are absent — `Requirements`, `Invariants`, `Behavioral Examples`, `Error Cases`, `Dependencies`, `Change Log`. A spec with valid frontmatter and the same body correctly flags every one of them.

## Why it matters

The exit code is right, so this will not let a bad spec through a gate. The **report** is what is wrong, and in two different ways:

- `✓ All source files exist` and `✓ All dependency specs exist` are **vacuous** — no `files:` or `depends_on:` was parsed, so zero of each were checked. Claiming success for an empty set reads as a pass.
- `✓ All required sections present` is **false**. Six are missing. This is not an unrun check reported as green; it is an incorrect answer to a check that did run.

Anyone fixing the frontmatter first and re-running will be surprised by six new errors that were always there. Worse, an agent consuming this output has no way to know the three green lines are worthless.

## Suggested fix

When frontmatter is invalid, do not emit affirmative results for checks that depend on it. Either skip them with an explicit marker — the `⊘ … skipped` shape the draft path already uses is exactly right — or run section validation against the default required set so the answer is true.

Found during RC1 stress testing. Verified by hand against a valid-frontmatter control.

Same family: CorvidLabs/spec-sync#547, #548, #549, #550, #551 — reporting success for work that did not happen.

0xLeif:
## Summary

CorvidLabs/spec-sync#553. A spec whose frontmatter cannot be parsed reported three checks as passing that never ran:

```
✗ Frontmatter invalid
  ✗ Missing or malformed YAML frontmatter (expected --- delimiters)
✓ All source files exist
✓ All required sections present
✓ All dependency specs exist
```

That spec's body contained two of the eight default required sections.

## Why this one is worse than the rest of the series

The other "green result for work that did not happen" findings are **vacuous** — they claim success over an empty set. This one is **false**. `✓ All required sections present` is a wrong answer to a question, not an unasked question reported as passed.

**Control, verified by hand:** the identical spec body with valid frontmatter reports **five** missing required sections. The sections really were missing; the tool really did say they were present.

## Root cause

All three lines infer success from the *absence of errors in their category*:

```rust
if file_errors.is_empty()    { "✓ All source files exist" }
if section_errors.is_empty() { "✓ All required sections present" }
if dep_errors.is_empty()     { "✓ All dependency specs exist" }
```

Unparseable frontmatter yields no `files:`, no `depends_on:`, and section validation never runs — so each category is empty for want of *input*, not for want of *problems*. Absence of evidence was being rendered as evidence of absence.

## After

```
✗ Frontmatter invalid
  ✗ Missing or malformed YAML frontmatter (expected --- delimiters)
⊘ Source file check skipped (frontmatter invalid)
⊘ Section validation skipped (frontmatter invalid)
⊘ Dependency check skipped (frontmatter invalid)
```

This reuses the `⊘ … skipped` wording the draft path already established, so there is one vocabulary for "this did not run" rather than two.

## Test Plan

- [x] Hand-verified before/after, plus the valid-frontmatter control proving the five sections are still reported once the check can run
- [x] Sandbox board green at `7bd6c0ac` — **28 pass, 0 fail, 0 skip**

## Notes

**Exit status is unchanged.** Invalid frontmatter was an error before and remains one, so this never let a bad spec through a gate — what was wrong was the report. The cost was borne by whoever read it: an author who fixes the frontmatter first is then surprised by errors that were always there, and an agent consuming the output has no way to know three of those lines were worthless.

A project whose specs all parse sees no difference at all.

Closes #553.


0xLeif:
Closing as incomplete — reopening as a complete fix.

A systematic sweep for this defect class found a **fourth** false claim this PR misses: `✓ All DB tables exist in schema`. Same fixture, four green lines rather than three:

```
✗ Frontmatter invalid
✓ All source files exist          ← fixed here
✓ All DB tables exist in schema   ← MISSED
✓ All required sections present   ← fixed here
✓ All dependency specs exist      ← fixed here
```

The db-tables line has the same defect behind a differently shaped guard — it tests whether the *project schema* has tables, not whether the spec's `db_tables:` could be read — so matching on the pattern of the other three stepped straight over it.

Shipping three quarters of a fix for "the tool reports success for checks that did not run" is the wrong trade. Replacement PR covers all four and rebases onto a main containing #557, which also resolves the duplicate `CHG-0110` allocation.

Issue #558, opened 2026-08-13
`stale` reports all specs up to date in a repository with zero commits


0xLeif:
Same class as CorvidLabs/spec-sync#547, #548, #550, #553: success inferred from the absence of evidence.

## Repro

A fresh `git init` with **no commits**, one source file, one spec:

```
$ specsync stale
  Result:    0/1 specs are stale
  ✓ All specs are up to date with their source files.
exit 0
```

Nothing was compared. Staleness is decided against git history, and there is none — no commit exists for the spec or its source to be newer or older than.

## Why it matters

`stale` is the command that answers "has my code moved on without its spec". Returning an affirmative "up to date" when it could not look is worse than returning nothing, because the answer is indistinguishable from a real one and a caller has no way to tell.

The zero-commit state is not exotic: it is exactly what `git init` + `specsync init` produces, which is the shape the quick start walks a new project through. It is also common in CI for a freshly created repository or a shallow checkout that fetched no history.

By contrast, `stale` outside a git repository entirely is handled correctly:

```
Error: Not a git repository — staleness detection requires git history.
exit 1
```

So the "no history to work with" case was considered — an unborn HEAD just was not recognised as an instance of it.

## Suggested fix

Detect an unborn HEAD and report it the way the no-git case is already reported, or report each spec as not-checked rather than as up-to-date. The `⊘ … skipped` vocabulary used elsewhere for "this did not run" fits.

Found while sweeping for the defect class after #553 turned out to have a fourth instance the original fix missed. Verified by hand against product main.

Issue #550, opened 2026-08-13
`deps` reports a malformed dependency graph as valid, exit 0


0xLeif:
Two commands, one tree, opposite verdicts — and the one specifically about dependencies is the wrong one.

## Repro

`specs/bad/bad.spec.md` with `depends_on: {oops` in frontmatter:

```
$ specsync check
  ✗ Frontmatter field `depends_on` must be a YAML list, got a mapping (offending line: `depends_on: {oops`)
exit 1

$ specsync deps
  Modules: 1  Edges: 0
  ✓ All dependency declarations are valid.
exit 0
```

The unparseable field is treated as an empty list and then affirmatively declared valid.

## Why it matters

`deps` is the command a user runs to ask specifically about dependency health. It is the *last* place that should paper over a dependency field it could not read. A project could carry a malformed `depends_on` indefinitely — `deps` green, edges silently missing from the graph and from the computed build order.

`deps` does work correctly on well-formed input: a genuinely missing dependency spec produces both the path error and the module error, exit 1. So this is specifically the unparseable-field path.

## Fix

`deps` should surface frontmatter parse errors rather than defaulting to an empty list, and must not print `✓ All dependency declarations are valid` for a spec whose declarations it failed to parse.

Found during RC1 stress testing. Verified by hand.

Issue #548, opened 2026-08-13
Plain `check` invents "requirements changed" warnings on a cold cache — every CI run


0xLeif:
On a fresh clone, `specsync check` reports a drift warning for every spec that has a `requirements.md`. There is no drift. There is no cache to have drifted from.

## Repro

Two runs, same fresh clone of `CorvidLabs/fledge` (`.specsync/hashes.json` is untracked, so a clone always starts cold):

```
run 1: exit 0 — 33 specs checked: 33 passed, 33 warning(s), 0 failed
       33 × "⚠ specs/<x>/<x>.spec.md: requirements changed — spec may need re-validation"
run 2: exit 0 — 0 × "requirements changed"
```

The warnings never come back until something actually changes.

## Cause

`classify_all_changes` compares each companion against `.specsync/hashes.json`. With no cache entry there is no baseline, and "unknown" is classified as "changed".

That is the right call for *deciding what to re-validate* — with no baseline you must re-check everything. It is the wrong call for *reporting to a person*, because you cannot truthfully say a file changed when you have nothing to compare it against.

## Why it matters

CI checks out fresh every time, so the cache is always cold there. **Every CI run of plain `check` on fledge prints 33 warnings that mean nothing.** That is how a team learns to ignore warnings — and the drift warning is one of the more valuable signals the tool has.

## Note for anyone reading the `--strict` difference

`check --strict` reports 0 of these where plain `check` reports 33, which looks alarming — as though `--strict` were suppressing drift. It is not. `src/commands/check.rs:273` empties the classification list under `force || strict || !spec_filters.is_empty()`, because those modes re-validate every spec and do not need to know which ones changed. The warnings that vanish are exactly these cold-cache artifacts. `--strict` produces the *more* correct output of the two.

Making `--strict` propagate these would be a regression, not a fix.

## Suggested fix

Emit the drift warning only when a cache entry exists and differs. Absent an entry, re-validate silently — the validation result is the signal, and it is a real one.

Found during RC1 stress testing. Verified by hand across two independent fresh clones.

0xLeif:
Fixed and merged in #552 (`fix(6.0): stop reporting success for checks that did not happen`), landed as `9240b1df`.

Left open by accident: that PR said `Closes #545, #547, #548`, and GitHub only honours the keyword for the first number in a list. Closing manually.

[...]

PR #561, opened 2026-08-13
fix(6.0): a tree with source and no specs must show its coverage and fail --strict (#560)

### LINKED ISSUE #560: `check --strict` exits 0 on a tree with source and zero specs, printing no coverage number
A project with real source and an empty `specs/` passes `check --strict` silently, and the log contains no number to notice it by.

## Repro

`src/auth/login.rs` with three exports, `specs/` empty:

```
$ specsync check --strict
No spec files found in .../specs/. Run `specsync generate` to scaffold specs.
exit 0

$ specsync coverage
File coverage: 0/1 (0%)
LOC coverage:  0/3 (0%)
exit 0
```

`check` prints the coverage footer in every other path. Here it is **omitted entirely**, so a CI log carries no figure at all — while `coverage` on the identical tree reports 0%.

## The code already intends otherwise

The branch that handles this carries a comment stating the requirement:

> No specs to validate — but a requested gate must still be evaluated against source coverage. Otherwise `check --require-coverage N`, `--enforcement enforce-new`, or **`--strict`** silently PASS in exactly the state they exist to catch: a project with source code but no specs.

`--require-coverage` and `--enforcement enforce-new` do gate. **`--strict` does not**, because `compute_exit_code` escalates *warnings* under `--strict`, and a project with no specs produces none. The code does not do what its comment says.

## Why it matters

This is the state right after `specsync init`, and the state of any repository whose `specs/` was deleted or mis-configured. A caller who asked for strict validation of a tree that was never measured is told it is clean.

`--format json` has the same gap: `specs_checked: 0` with `passed: true` is indistinguishable from an empty project.

## Suggested fix

1. Print the coverage line in this path unconditionally, not only when a gate already failed.
2. Gate under `--strict` when the tree has source files — confined to that, so an empty project, or one whose specs simply have not been generated yet, still exits 0.
3. Include `total_source_files` and `coverage_percent` in the JSON payload so a consumer can tell the two cases apart.

Found by sweeping for affirmative claims after #553 turned out to have a fourth instance the first fix missed. Same class as #546, #547, #548, #550, #553, #558.

0xLeif:
## Summary

CorvidLabs/spec-sync#560. A project with real source and an empty `specs/` passed `check --strict` silently, and the log contained no number to notice it by:

```
$ specsync check --strict
No spec files found in .../specs/. Run `specsync generate` to scaffold specs.
exit 0

$ specsync coverage
File coverage: 0/1 (0%)
LOC coverage:  0/3 (0%)
exit 0
```

`check` prints the coverage footer in every other path. Here it was omitted entirely.

## The code already intended otherwise

The branch handling this carries a comment stating the requirement outright:

> No specs to validate — but a requested gate must still be evaluated against source coverage. Otherwise `check --require-coverage N`, `--enforcement enforce-new`, or **`--strict`** silently PASS in exactly the state they exist to catch: a project with source code but no specs.

`--require-coverage` and `--enforcement enforce-new` do gate. **`--strict` did not** — it escalates *warnings*, and a project with no specs produces none. The code did not do what its own comment said, which is worth noting: reading the comment and believing it would have closed this as already-handled.

## After

| tree | before | after |
|---|---|---|
| source present, no specs — bare `check` | no coverage figure | `File coverage: 0/1 (0%)`, exit 0 |
| source present, no specs — `--strict` | exit 0 | **exit 1** |
| source present, no specs — JSON | `specs_checked: 0`, nothing else | `total_source_files: 1`, `coverage_percent: 0` |
| **no source, no specs** | exit 0 | **exit 0, unchanged** |
| any project with ≥1 spec | — | unchanged |

## Test Plan

- [x] Four fixtures hand-verified, including the control

**The risk here was over-reaching, not under-reaching.** `check` failing on a freshly initialized repository is precisely the defect #544 fixed, so a gate firing whenever specs are absent would have reintroduced it. The gate is confined to trees that actually contain source, and the no-source control is the load-bearing assertion — without it this change is indistinguishable from one that simply fails harder.

## Adopter impact

Projects adopting spec-sync will see `check --strict` fail until the first spec exists. That is the intended signal: strict validation of a tree that was never measured should not report clean. Bare `check` is unchanged and remains the right command during setup.

## Notes

Sixth instance this cycle of one class — a category empty for want of *input*, read as want of *problems* (#546, #547, #548, #550, #553, #560). This one was found by sweeping for affirmative claims rather than from a report; the same sweep produced #558 and caught a fourth missed line inside #553's original fix.

Closes #560.


0xLeif:
Superseded by #563.

This branch was cut from main before #559 merged, so both changes were allocated `CHG-0111` and the sequence ledger conflicted alongside the CHANGELOG. Rather than hand-write an `acknowledged_collisions` entry — guessing at ledger semantics the tool itself validates — the change was reapplied on the current main and re-run through the lifecycle, which allocated `CHG-0112` cleanly.

Same 24-line source change, re-verified: 2210 unit + 331 integration, and both fixtures still hold (gates with source present, stays clean on an empty project).

PR #579, opened 2026-08-14
fix(6.0): a config file that cannot be loaded must refuse to run, not report success (#570)

### LINKED ISSUE #570: An unparseable config.toml silently disables every configured rule and the run reports success
One missing bracket in `.specsync/config.toml` turns a failing project green. The diagnostic goes to **stderr only**, so a CI job capturing stdout sees a clean run.

## Repro — two trees, byte-identical except one character

`.specsync/config.toml` sets `required_sections` to the seven built-ins plus `"Threat Model"`. The spec satisfies the built-ins and has no Threat Model section.

**Control, valid config:**

```
$ specsync check --root ok
  ✗ Missing required section: ## Threat Model (check config: .../ok/.specsync/config.toml)
    -> Add `## Threat Model` heading to the spec body
exit 1
```

**Same tree, closing `]` deleted:**

```
$ specsync check --root broken        # stdout, as CI captures it
  ✓ Frontmatter valid
  ✓ All source files exist
  ✓ All required sections present
  ✓ 1/1 exports documented
1 specs checked: 1 passed, 0 warning(s), 0 failed
exit 0
```

The warning exists, but only on stderr:

```
Warning: failed to parse config file .../broken/.specsync/config.toml: TOML parse error at line 3, column 1 ... ; using built-in defaults
```

## It survives every escape hatch

- `check --strict --force --json` → exit 0, and the payload is silent: `{"errors": [], "passed": true, "specs_checked": 1, "warnings": []}`
- `score --root ok` → `96/100` (docked for the missing section); `score --root broken` → **`100/100`**
- `score --min-score 90` exits 0 in both

## Why this is the worst instance of the class

Ten defects fixed recently share one shape: *a category is empty for want of **input**, and the code reads that as want of **problems**.* This is that bug at the **configuration layer**, so it does not disable one check — it disables every rule the project configured beyond the built-in defaults, simultaneously, and reports success.

`✓ All required sections present` is a claim about a section list that was thrown away.

The failure is also silent in the direction that matters: a project adds `required_sections`, `[rules]` thresholds, or a custom `specs_dir` precisely because the defaults are not enough. A typo in that file returns them to the defaults they rejected, and tells them everything is fine.

## Precision

The run reaches exit 0 only because the spec independently satisfies the built-in defaults. The defect is that **configured-beyond-default checks vanish without failing the run**. This repositorys own `.specsync/config.toml` happens to set `required_sections` equal to the seven defaults, so that field is coincidentally harmless here — `[rules]` thresholds and any added section are not.

## Suggested fix

A config file that exists and cannot be parsed should be an **error**, not a fallback. The project asked for those rules; silently substituting different ones is never the safe reading. If a fallback must stay for compatibility, it belongs behind an explicit flag, and the run must not report success while it is active.

Compare `deps`, which now correctly fails closed on malformed frontmatter (#550), and `check`, which reports `⊘ … skipped` rather than `✓` when frontmatter cannot be parsed (#553). Same principle, one layer up.

Found by an agent probe of commands with no drill coverage; verified by hand against a valid-config control.

0xLeif:
## Summary

CorvidLabs/spec-sync#570. One missing bracket in `.specsync/config.toml` turned a failing project green.

Two trees, byte-identical except one character. The config sets `required_sections` to the seven built-ins plus `"Threat Model"`; the spec has no Threat Model section.

```
$ specsync check --root ok        # valid config
  ✗ Missing required section: ## Threat Model
exit 1

$ specsync check --root broken    # one `]` deleted — stdout, as CI captures it
  ✓ All required sections present
1 specs checked: 1 passed, 0 warning(s), 0 failed
exit 0
```

The warning existed — on **stderr only**.

## Why this is the worst instance of the series

Ten prior fixes share one shape: a category empty for want of *input*, read as want of *problems*. Each disabled a single check. This is that bug at the **configuration layer**, so it disables every rule the project configured — `required_sections`, `[rules]` thresholds, `exclude_patterns` — simultaneously, from one typo.

It survived every escape hatch:

| attempt | result |
|---|---|
| `--strict` | exit 0 |
| `--force` | exit 0 |
| `--json` | `"passed": true`, `"errors": []` |
| `score` | **rose** 96 → 100/100 |
| `score --min-score 90` | exit 0 |

The score movement is the clearest evidence of the mechanism: the project lost the very rule it was failing, so its grade improved.

`✓ All required sections present` is a claim about a section list that had been thrown away.

## The intent was already written down

The fallback sites carry comments saying a present-but-unreadable config *"must fail loud rather than silently downgrade enforcement to defaults"*. They then warn and return the defaults. Same shape as #560: the requirement was understood, recorded, and not delivered.

## Two implementation notes worth reading

**Four fallback sites, two shapes.** Two in `config.rs` handle an unreadable file; the *parse* failure — which is what #570 actually reproduces through — is a separate pair in `validator.rs` reached during retained discovery. A fix pattern-matched on the first shape compiles, looks right, and leaves the reported defect untouched. My first attempt rewrote zero of the relevant sites.

**One choke point, not thirty-six.** Gating in `compute_exit_code`/`exit_with_status` meant threading a parameter through **36 call sites across 8 files**, and it fails open — any command omitting the argument silently keeps the old behaviour. The guard sits in `load_and_discover` instead, which every spec-reading command uses. It also covers `generate` and `scaffold`, which would otherwise scaffold into a `specs_dir` the project never configured.

## Test Plan

| fixture | before | after |
|---|---|---|
| config exists, one `]` deleted | exit 0, `✓ All required sections present` | **exit 1**, refuses and names the file |
| **control** — valid config | exit 1, reports the missing section | unchanged |
| **control** — no config file at all | exit 0 | **exit 0**, unchanged |

The third row confines the change. The refusal fires only when a file **exists and could not be loaded** — never on its absence, which is a legitimate configuration and the state every new project starts in.

## Notes

Refusal is a hard exit rather than a warning escalated under `--strict`. `--strict` means "treat warnings as errors", and this is not a warning: every configured rule is absent. Leaving it opt-in would leave the default path — the one CI most often uses — reporting success.

Found by an agent probing the commands that have **no drill coverage at all**, rather than from a bug report. Eleven rounds of fixing individual commands had not surfaced it, because it sits at the layer beneath them.

Closes #570.


0xLeif:
Closing — replaced by a clean rebuild, not abandoned.

This branch was cut from `main` while #575 was still open, so both changes allocated `CHG-0117` and the sequence ledger conflicted. Rather than hand-write an `acknowledged_collisions` entry — guessing at ledger semantics the tool itself validates — the change is being reapplied on the current main and re-run through the lifecycle.

Worth recording what the rebuild caught: the diff I extracted spanned a base from *before* #575 merged, so replaying it onto main silently **reverted** the coverage fix in `src/output.rs` and `specs/output/`. `git apply` succeeded, the build passed, and every config fixture behaved correctly — every signal said fine. It only surfaced by counting occurrences of the coverage fix (0 on the branch, 1 on main) and restoring those files.

Same source change, same three verified behaviours: broken config exits 1, valid config unchanged, no config at all still exits 0 on built-in defaults. Replacement PR follows once the suite is green.

Issue #637, opened 2026-08-18
ci: no check verifies that a check-run name a workflow waits on is actually produced — the mechanism behind #635


0xLeif:
## Every check-run name a workflow waits on should be proven to have a producer

#635 happened because #499 deleted a workflow that emitted a check run, and left the workflow
that waits for it. Nothing detected that for sixteen days, because the waiting workflow's first
execution is the release itself.

This is mechanically checkable and cheap. `release.yml` selects check runs by name:

```
.name == "specsync"
.name == "SpecSync archive binding"
.name == "SpecSync release candidate"
```

Two of those three resolve to a producer in `trust.yml`. The third resolved to nothing. A
script that extracts the waited-on names and asserts each appears in some other workflow or
script under `.github/` fails on `main` today and passes after #635's fix — verified by hand
against both files in the same tree:

```
BEFORE  NO PRODUCER for 'SpecSync archive binding' — the wait can never be satisfied   FAIL
AFTER   all waited-on names have producers                                             PASS
```

It belongs in `.github/scripts/` next to `validate-workflow-runtime-pins.py`, which is the same
family of static workflow validation, with a `test-` companion per the existing convention, and
wired into the `spec-check` or `validate-action` lane so it runs on every PR.

## Why it is filed rather than included in #635's fix

#635's change declared `affected_paths: .github/workflows/release.yml` at its interview, and
delivery scope cannot be widened in flight (#542). Adding `.github/scripts/` would have
required abandoning and re-opening that change.

That is a small, concrete instance of #542's cost: the blast radius of a fix became clear only
while verifying it, which is after the point where scope froze. Worth noting on that issue as
a real example rather than a hypothetical.

## Scope

- extract every `.name == "…"` used to select a check run, across all workflows, not just
  `release.yml`
- assert each has a producer under `.github/`
- fail with the name and the waiting file when one does not
- a `test-` companion covering: a name with a producer (passes), a name without (fails), and a
  workflow that waits on nothing (passes, and does not pass vacuously)

Issue #626, opened 2026-08-17
Required CI gate counts skipped as pass, so a PR can change src/ with test, fmt and coverage never run


0xLeif:
## What

`.github/workflows/ci.yml:548-558`:

```yaml
- name: Require every selected gate
  env:
    RESULTS: ${{ join(needs.*.result, ' ') }}
  run: |
    for result in $RESULTS; do
      case "$result" in
        success|skipped) ;;
        *) echo "Selected CI gate ended with: $result" >&2; exit 1 ;;
      esac
    done
```

A skipped job satisfies the aggregate. That is correct when a job is skipped because it is genuinely irrelevant, and wrong when the path classifier gets the classification wrong — the gate cannot tell the two apart.

## It has already happened

**PR #567** (merged as `60360ed2`) modified **`src/commands/check.rs`**. Its checks:

```
pass      Required CI gate          <-- passed
pass      SpecSync implementation ready
pass      Lifecycle gate / preflight / trust / CodeQL / Analyze
skipping  test
skipping  fmt
skipping  coverage
skipping  audit
skipping  spec-check
skipping  SpecSync scoped review
```

A product source change merged to `main` with its test, format, coverage, audit and spec-validation gates never executed, and the required aggregate reported green.

The likely cause is the classifier: 19 of the PR's 20 files are under `.specsync/archive/changes/…`, so it was treated as a lifecycle/archive tip and the product lane was skipped — but the PR also carried a real `src/` change. Archive tips legitimately skip product checks and reuse trust from a green product parent; the classification is only safe if "archive tip" and "touches src/" are mutually exclusive, and here they were not.

## Why it matters

This is the aggregate check the branch relies on. Combined with #624 — where `coverage` fails on every push to `main`, so the same gate is red for an unrelated reason — `Required CI gate` currently cannot distinguish any of: everything ran and passed; nothing ran at all; coverage infrastructure died. Three very different states, one signal.

It is the release's defect class in the CI itself: the lane is empty for want of *selected jobs*, and the gate reads empty as clean.

## Fix

Two independent changes, both small:

1. **Never skip a product gate when the diff touches product paths.** If `Classify changed paths` sees any `src/**`, `tests/**`, or `Cargo.toml` entry, the product lane must run regardless of what else is in the diff. An archive tip that also touches `src/` is not an archive tip.
[...]

Issue #542, opened 2026-08-12
feat(change): delivery scope cannot be widened in flight, but blast radius is only knowable after compiling and testing


0xLeif:
## Summary

Delivery scope (`affected_paths` / `affected_specs`) is frozen when the interview is answered, and there is no supported way to widen it afterwards. `change answer` requires `draft`:

```
error: cannot answer interview questions while CHG-0104-… is verifying; expected draft
```

and nothing transitions a change back to `draft`. `change correct` handles `public_contract` / `architecture_risk` only; `change correct-owner` repairs *acceptance* owners, not the declared path set.

The problem is that **scope must be declared before the change can be compiled or tested** — which is before its true blast radius is knowable.

## Evidence: one small change, three scoping attempts

Severing `specsync check` from the trust layer (a ~30-line edit to two command files) had this actual radius:

| Discovered at | What | Could it have been known earlier? |
|---|---|---|
| design time | `src/commands/check.rs`, `src/commands/comment.rs` | yes |
| **compile time** | `src/change.rs` — severing `comment.rs` orphaned `check_project_quiet`, its only caller, leaving `ConfiguredCommandOutput` with one live variant threaded through three call sites | no — `cargo clippy -- -D warnings` is what surfaces it |
| **test time** | `tests/integration/check.rs`, `tests/integration/comment.rs` — three tests asserting the removed behavior | no — requires running the suite |

Neither later discovery could be acted on: both touch meaningful paths outside the approved scope, so `check` refuses with `meaningful changed paths are not covered by an active change`, and scope cannot be widened. Each time, the only route was to discard the workspace and recreate it — losing the approval, the artifacts' identity, and a full verification cycle (~18 minutes each, because `src/change.rs` is in the hardcoded strict tier).

The third attempt only succeeded because the full test suite was run *before* creating the change, purely to establish the radius. That is a workaround, not a workflow.

## Why the existing escape hatches do not cover it

- `change correct` — interview metadata only (`public_contract`, `architecture_risk`).
- `change correct-owner` — appends an acceptance-manifest owner for a path *already in delivery scope*. It cannot add a path.
- `change reopen` — returns an archived/accepted change to `verifying`, not to `draft`, so the interview is still closed.
- Recreating — loses the change identity and every prior gate, and burns a verification cycle.

## Proposed

A supported scope-widening operation on a pre-terminal change, e.g.

```
specsync change scope <id> --add-path <PATH> --add-spec <MODULE> --reason "<why>"
[...]

PR #638, opened 2026-08-18
ci(release): the release lane must not gate on a check no workflow produces (#635)

### LINKED ISSUE #635: ci(release): the release lane gates on `SpecSync archive binding`, which no workflow has produced since #499 — v6.0.0 cannot be tagged
## The release lane waits for a check that nothing produces

`release.yml:173` gates every release on a check run named **`SpecSync archive binding`**. No
workflow in this repository emits it:

```
$ grep -rn 'SpecSync archive binding' .github/workflows/ci.yml .github/workflows/trust.yml .github/scripts/
(no matches — the only references are the consumers in release.yml)
```

Its producer was `post-merge-archive.yml`, deleted in #499. Nothing replaced it.

## Measured

Check runs actually present on `200add8f` (current `main` tip):

```
Analyze (actions), Analyze (javascript-typescript), Analyze (rust), Build Astro Site,
Classify changed paths, Deploy to GitHub Pages, Lifecycle gate, Lifecycle preflight,
Packaged GitHub Action consumer, Record attestation, Required CI gate,
SpecSync implementation ready, SpecSync scoped review, audit, coverage, fmt, site,
spec-check, test, trust, validate-action, vscode-extension
```

`SpecSync archive binding` is not among them, and never has been on any commit I checked.

## Consequence

The wait loop is unconditional — no `if`, no `continue-on-error`:

```bash
for attempt in $(seq 1 12); do
  ...
  binding_status="…| select(.name == "SpecSync archive binding") …| if . == null then "missing" …"
  ...
  sleep 10
done
if [ "$binding_status" != "completed/success" ]; then
  echo "::error::Release commit has no structurally valid SpecSync archive binding check"
  exit 1
fi
```

`binding_status` is `"missing"` on all twelve attempts, so the job spends 120s and then exits 1
— and the job carries `timeout-minutes: 5`, so on a slow API it dies of timeout first. Either
way **`v6.0.0` cannot be tagged.**

Everything downstream of it in `release.yml` — the digest binding at :381-468, artifact
publication — has therefore never executed. The lane has not run once since #499.

## Why it went unnoticed

Same shape as the rest of this release: *tested component reached through untested wiring*.
The validation logic below the wait (:381-468) is careful and detailed — it parses the
`external_id`, checks the app slug and id, pins the `details_url` prefix, cross-checks the PR
binding two ways. All of it reads as a working system. None of it is reachable, because the
input it validates is never produced.

There is no test, no dry run, and no lane that exercises `release.yml` before a tag, so the
first time anyone learns this is when they push the `v6.0.0` tag.

## Fix — one of

1. **Restore the producer.** Re-add the post-merge job that emits `SpecSync archive binding`
   with the `external_id` shape the validator already expects
   (`specsync-archive:CHG-NNNN-…:<64 hex>`). This keeps the binding guarantee that #499's
   deletion silently dropped.
2. **Drop the gate.** If the binding is no longer the intended mechanism, delete the wait and
   the validator with it, rather than leaving 300 lines that validate an input nothing sends.

(1) preserves the property; (2) is honest about it. What is not acceptable is the current
state, where the release lane claims to enforce a binding it can never observe.

Either way this needs a way to exercise `release.yml` short of tagging — a `workflow_dispatch`
dry-run path, at minimum — or the next deletion of a producer will land exactly the same way.

## Blocking

Hard blocker for the 6.0 tag. Not a product defect: the binary is unaffected. It is the reason
the internal RC cannot be cut today even with every product gate green.

0xLeif:
## Summary

`release.yml`'s `validate` job waited for a check run named **`SpecSync archive binding`** and
then validated it in ~430 lines of embedded Python. Nothing in this repository produces that
check run.

Its producer was `.github/workflows/post-merge-archive.yml`, deleted in `802ca13b` (#499).
**#499 removed the producer and left the consumer.**

The wait carried no `if` and no `continue-on-error`, so `binding_status` stayed `missing`
through all twelve attempts and the job exited 1 — or hit its `timeout-minutes: 5` first.

## This blocked the RC, not only the final tag

`validate` is the only job in the lane with no mode guard:

| job | guard |
|---|---|
| `validate` | **none** |
| `qualify`, `record-qualification` | `mode == 'qualify'` |
| `authorize-release`, `promote`, `build` | `mode == 'promote'` |

The workflow's only triggers are an RC tag push (`mode=qualify`) and a dispatch
(`mode=promote`). Both run `validate`. So pushing `v6.0.0-rc.1` failed before `qualify`
started — the internal RC was as blocked as `v6.0.0` itself.

## Deleting the gate, not restoring the producer

#499's stated purpose was removing ~7,257 lines of Python, bash and YAML that re-derived the
SDD lifecycle from Git commit topology, on the grounds that SpecSync implements it in tested
Rust that ships to users. Its commit message names a bug living in this very file:

> post-merge-archive.yml counted a change package's `deltas/` subdirectory as a second archive
> root, so #494's archive never bound to its merge commit

Restoring it would reinstate what #499 removed, including a `deltas/` bug in the same subject
this release just fixed on the product side (#536).

### What is genuinely lost

Real properties go with the block: that the bound pull request was merged, that the archive is
`workflow_version=2`, that its finalization evidence validates, that the PR has an exact
implementation head. They are unreachable today, but they are not nothing. After this change
the archive-to-merge-commit binding is enforced **solely by SpecSync itself**, in `change ship`
and archive validation — which is where #499 put the lifecycle. The release lane stops
re-deriving it from Git topology.

### What survives

`validate` keeps the three checks that never depended on the binding: tag version equals the
`Cargo.toml` package version, the checkout is the resolved candidate, the candidate is an
ancestor of `origin/main`. `GH_TOKEN`, `REPOSITORY` and `SERVER_URL` drop out of its `env`
because only the deleted half used them.

## Making the lane runnable without a tag

The root cause of this sitting undetected for sixteen days is that the lane's first execution
*is* the release. A `dry_run` dispatch input now resolves `mode=dry-run`: `validate` runs
because it has no guard, and every other job is already gated on `qualify` or `promote`, so all
of them skip and nothing is tagged or published.

```
qualify   resolve validate qualify record-qualification
promote   resolve validate authorize-release promote build release
dry-run   resolve validate
```

An unrecognized value is **rejected**, not defaulted:

```bash
case "${REQUESTED_DRY_RUN:-false}" in
  true) mode="dry-run" ;;
  false|"") mode="promote" ;;
  *) echo "::error::dry_run must be true or false, got '${REQUESTED_DRY_RUN}'"; exit 1 ;;
esac
```

The first draft used `[[ "$X" == "true" ]]` with an `else` to `promote`, so `dry_run=TRUE` —
reachable via `gh workflow run -f dry_run=TRUE` — would have silently run a **real promotion**
for someone who asked for a dry run. The simulation below caught it. An unrecognized input
quietly reading as a safe default is exactly the defect class this release is about.

## Verification

This is Actions configuration and cannot be executed locally. The honest statement of the limit
is that the lane is still unexercised — that is the defect being fixed, and `dry_run` is what
finally makes exercising it possible.

**The discriminating test.** Every check-run name `release.yml` selects on must have a producer
under `.github/`. Run against `origin/main`'s file and this one, in the same tree:

```
════ BEFORE (origin/main 8fad38d4) ════
  waits on: specsync, SpecSync archive binding, SpecSync release candidate
  PRODUCER FOUND for 'specsync': trust.yml ci.yml …
  NO PRODUCER for 'SpecSync archive binding' — the wait can never be satisfied
  PRODUCER FOUND for 'SpecSync release candidate': trust.yml
  verdict: FAIL   rc=1

════ AFTER (this change) ════
  waits on: specsync, SpecSync release candidate
  PRODUCER FOUND for 'specsync': trust.yml ci.yml …
  PRODUCER FOUND for 'SpecSync release candidate': trust.yml
  verdict: PASS   rc=0
```

The two surviving names are the control: the check does not pass because it stopped looking.
`release.yml` still waits on two check runs, both resolving to a real producer in `trust.yml`.
Only the third had none.

**Mode simulation**, the `resolve` branch extracted verbatim:

```
PASS  RC tag push                          -> qualify
PASS  dispatch, dry_run unset              -> promote
PASS  dispatch, dry_run=false (control)    -> promote
PASS  dispatch, dry_run=true               -> dry-run
PASS  dispatch, dry_run=TRUE fails closed  -> REJECTED
PASS  dispatch, dry_run=yes fails closed   -> REJECTED
PASS  branch push still unsupported        -> UNSUPPORTED
PASS  final tag push still unsupported     -> UNSUPPORTED
```

**Static:** `yaml.safe_load` parses; `actionlint` clean; 8 jobs with unchanged names and
dependency edges; no `needs.validate.outputs` anywhere (the deleted step had no `id` and the
job exposes no outputs); no remaining reference to `archive binding`, `archive-binding` or
`specsync-archive`.

## A second scope escape, found by CI

The first CI run failed. `.github/scripts/test-validate-release-candidate.py` reads
`release.yml` as **text** and anchored on `export CHECKS_FILE="$checks_file"`, a line inside the
deleted block:

```
ERROR: test_release_reconstruction_requires_actual_pull_request_event
ValueError: substring not found
FAILED (errors=1)
```

`Classify changed paths` died, and because every job depends on it, ten gates reported
`skipping` and both required gates failed. That cascade is worth naming: a single orphaned
string anchor converted ten gates into "did not run", which reads the same as "did not object" —
the same shape as #626.

Rather than fix the one test CI happened to name, I resolved **all twenty** `release.index(...)`
anchors in that file against the new workflow:

```
MISSING ANCHOR: '          export CHECKS_FILE="$checks_file"'
1 of 20 orphaned
```

The nineteen survivors are the control: had all twenty failed, the conclusion would be that the
file no longer describes the workflow at all, not that one block was removed. Also checked the
two assertions near the `resolve` job, since this change edits it —
`assertNotIn('mode="release"')` and `assertNotIn("needs.resolve.outputs.mode == 'release'")`
both still hold, because the added mode is `dry-run`.

That test is now removed, with a comment where it stood recording what it asserted and why its
subject is gone. Suite: `Ran 49 tests … OK`.

**It landed as a second change (CHG-0151).** The test file is outside CHG-0150's declared
`affected_paths`, CHG-0150 was already archived, and `change audit` refused the commit and named
the remedy itself. So this PR carries two sequential changes — the fix, and the consequence of
the fix that only CI could reveal.

## Known gap, filed rather than smuggled in

The producer/consumer check above is run by hand. Making it permanent means adding
`.github/scripts/`, outside this change's declared `affected_paths`, and delivery scope cannot
be widened after the interview (#542). Filed as #637 — and noted there as a small concrete
instance of #542's cost, since the blast radius only became clear while verifying.

## Incidental confirmation of #536

This change is `--no-spec-change`, so its `deltas/` directory was empty — precisely the case
that produced the husk fixed in #636 one commit earlier. Its archive package came out with the
directory pruned and zero empty directories:

```
empty dirs in .specsync/archive/changes/2026-08-18-CHG-0150-…:  none
deltas/:                                                        absent (pruned)
```

On the previous binary this package would have been a landmine for the next branch switch. Not
evidence for this change, but a real dogfooding datapoint for the one before it.

## Test Plan

- [x] Producer/consumer check fails on `origin/main`'s file, passes on this one, two controls green in both
- [x] Eight-case mode simulation including two fail-closed rows
- [x] `qualify` and `promote` job graphs unchanged; `dry-run` reaches nothing that tags or publishes
- [x] `actionlint` clean, YAML parses, no dangling `needs.validate.outputs`

Fixes #635.

https://claude.ai/code/session_01DgYAvsQM6P9fKxDotnDuzP


0xLeif:
## On the red CodeQL check

CodeQL reports "1 new alert including 1 high severity security vulnerability". It is **not new**
— it is a pre-existing alert re-reported at a different line because this change moved the step
it points at.

| | rule | severity | path | message |
|---|---|---|---|---|
| `main`, alert #62 | `actions/cache-poisoning/poisonable-step` | high | `release.yml:173` | …from `needs.resolve.outputs.candidate_sha`. (workflow_dispatch) |
| this PR | `actions/cache-poisoning/poisonable-step` | high | `release.yml:210` | …from `needs.resolve.outputs.candidate_sha`. (workflow_dispatch) |

Same rule, same message, same step — `main:173` is `Validate candidate source and merged
archive binding`, which this change renames to `Validate candidate source` and the explanatory
comment block pushes down to 210. CodeQL keys alerts by location, so a moved line reads as a new
finding.

Three of these are already open on `main` (`release.yml:173`, `:848`, `:1064`). None is
introduced here, and this change does not widen the exposure: before it, `workflow_dispatch`
always resolved `mode=promote`; after it, dispatch resolves either `promote` or `dry-run`, and
`dry-run` reaches strictly fewer jobs.

CodeQL is not a required check — `Required CI gate` is the only one, and it passed along with
the other 17.

I have **not** waved the finding away. It is real, and while confirming the relocation I found
that the ordering inside this very step makes it worse than it needs to be: `cargo metadata`
runs on the checked-out candidate *before* the ancestor check validates that candidate. Filed
separately with that detail rather than folded in here, since this change is archived and scoped
to the deletion.

PR #664, opened 2026-08-20
fix(6.0): a change without an ordinal must not brick the workspace

0xLeif:
A landmine shipped by CHG-0162 (`65755ac7`), found while scoping the sequence-ledger retirement rather than by anyone hitting it.

## Reproduced end to end

One hand-made slug-only change directory, against the shipped binary:

```
$ specsync change audit --strict
error: invalid change ID `a-slug-only-change`

$ specsync change new "can I still work"
error: invalid change ID `a-slug-only-change`

$ specsync change status
CHG-0001-a-normal-change        draft   next: run `specsync change answer …`
CHG-0002-another-normal-change  draft   next: run `specsync change answer …`
```

**No further change can be created at all** — and `change status` lists everything as healthy and says nothing about it.

## Cause

CHG-0162 relaxed `validate_change_id` so an identity without the `CHG-NNNN` prefix is legal, which is the entire point of the identity work. What it did not do is check whether anything downstream still required the ordinal.

`located_change_sequences` (`src/change.rs:1958`) did:

```rust
let sequence = change_sequence(&record.id)
    .ok_or_else(|| format!("invalid change ID `{}`", record.id.escape_default()))?;
```

A slug-only record now loads successfully and reaches that `?`. `validate_change_sequences` sits on both the `change audit` and `change new` paths, so one such directory is fatal to both. Archive manifest recomputation dies on it too — `historical_sequence_ledger_acceptance_content` calls the same function.

The silence is its own defect: `sequence_ledger_freeze_next_action` (`:2023`) pattern-matches known error strings and ends `Err(_) => None`, so an unrecognised error yields a clean next-action. **The project looks fine and is bricked.**

## Fix

The ordinal is read at that site for exactly one purpose — detecting two changes that claimed the same number. An ID with no number cannot collide, so it does not participate:

```rust
let Some(sequence) = change_sequence(&record.id) else {
    continue;
};
```

This is the sibling-site pattern again: the fix landed where the report pointed (`validate_change_id`) and a second validator with the same job survived beside it, **stricter by accident rather than by decision**.

## Measured, before and after

Same scratch repository, one slug-only directory:

```
before   error: invalid change ID `a-slug-only-change`      (change new dead)
after    Auditing active changes (3)…                       (reports the real problem)
control  error: duplicate numeric change sequence CHG-0001  (still refused)
```

## Discrimination

Against a separate checkout of `origin/main`:

| Test | `origin/main` | here |
|---|---|---|
| `a_workspace_holding_an_ordinal_free_change_still_enumerates` | **FAIL** — `invalid change ID` | pass |
| `two_changes_claiming_one_ordinal_are_still_refused` | pass | pass ← control |

The control is what stops the fix being "skip everything": collision detection has to survive, and it does.

## Reachability today

Nobody can hit this accidentally on `main` right now — `change new` still mints `CHG-NNNN-slug`, so it takes a hand-made directory or a future binary. It becomes reachable the moment the allocator mints slug-only IDs, which is the next change. Landing it first is deliberate.

## Test plan

- [x] `cargo test` green
- [x] `cargo fmt --check` clean
- [x] `specsync check --strict` and `change audit --strict` both rc=0
- [x] Reproduced end to end against the shipped binary before fixing
- [x] Discrimination measured against a separate checkout


0xLeif:
Closing in favour of step 5, which contains a **corrected** version of this fix.

The change here is too permissive. It does:

```rust
let Some(sequence) = change_sequence(&record.id) else { continue; };
```

`change_sequence` returns `None` for two different situations, and this conflates them:

| ID | claims an ordinal? | `change_sequence` | should |
|---|---|---|---|
| `a-slug-only` | no | `None` | skip — nothing to collide |
| `CHG-16-narrow` | **yes**, malformed | `None` | **fail closed** |
| `CHG-00016-wide` | **yes**, malformed | `None` | **fail closed** |

Those last two hard-error on `main` today. This PR would silently drop them, taking them out of the acknowledged-collision ID-set check that guards the archived collision members — a small regression riding along with a real fix.

Step 5 distinguishes "claims none" from "claims one badly", and adds the thing this PR does not: an **ordinal-independent duplicate-identity gate**. The ordinal was the repository's only guarantee that two changes could not share a name, and removing it without a replacement means two same-named archives report `✓ audit passed`.

The brick this fixes is not reachable on `main` today — `change new` still mints `CHG-NNNN-slug`, so it takes a hand-made directory. It becomes reachable exactly when the allocator starts minting slug-only IDs, which is step 5. Fixing it there, correctly, rather than here, partially.

CHG-0164 stays in the archive as the record of the investigation.

